Issue #16, opened 2026-05-22
Aspect-aware image cropping in Helpers::resizeImage / |resizer — brainstorm


parisek:
## Problem

Components that crop images to a fixed aspect ratio (e.g. `article-gallery`'s `image_pair` layout cropping to 4:3 landscape) lose significant visible area when the editor uploads a portrait or square source.

Concrete example from a downstream consumer (pm-a's `article-gallery.twig`, `image_pair` layout):

```twig
{{ component_picture({
    image: item.image|resizer(
        ['960', '720', '1280', 'crop'],
        ['660', '495', '1024', 'crop'],
        ['480', '360', '', 'crop'],
    ),
    ...
}) }}
```

These tuples crop **every** source to a 4:3 landscape frame. Outcomes:

| Source orientation | Source dimensions | After crop to 960×720 | Visible loss |
| --- | --- | --- | --- |
| Landscape 16:9 | 1920 × 1080 | 960 × 720 | 0 % — fits naturally |
| **Portrait 9:16** | **1080 × 1920** | **960 × 720 (centre band)** | **≈ 70 % of subject height** |
| Square 1:1 | 1080 × 1080 | 960 × 720 (centre band) | ≈ 33 % of height |

Editors can't always know the layout's crop direction when uploading. The component should adapt to source orientation rather than imposing a single crop.

## Current workaround (per-component, no kit support)

I implemented variant A (inline Twig conditional) in pm-a — works correctly but duplicates intent across every component that crops:

```twig
{% set is_landscape = (item.image[0].width|default(1)) >= (item.image[0].height|default(1)) %}
{% set sized_image = is_landscape
    ? item.image|resizer(['960', '720', '1280', 'crop'], ['660', '495', '1024', 'crop'], ['480', '360', '', 'crop'])
    : item.image|resizer(['720', '960', '1280', 'crop'], ['495', '660', '1024', 'crop'], ['360', '480', '', 'crop'])
%}
```

[Reference component fix in pm-a article-gallery.twig](https://github.com/portadesign/pm-a/blob/feature/styleguide/wp-content/themes/pm-a/static/templates/component/article-gallery/article-gallery.twig#L96-L110)

The downside of solving this per-component: every gallery / teaser / card with a fixed-crop tuple needs the same 4-line conditional. Square-source case is also unhandled (`>=` falls through to landscape branch — gives 33 % visible loss on square uploads).

## Proposal — solve in timber-kit's Helpers / Twig API

[...]

PR #18, opened 2026-05-22
feat: add |resizer_aspect Twig filter for orientation-aware cropping (#16)

parisek:
## Summary
- Implements parisek/timber-kit#16 — new `|resizer_aspect` Twig filter that classifies a source image's orientation (landscape / portrait / square, ±10 % tolerance around 1:1) and routes to a per-orientation tuple set, so consuming components can drop the inline `image.width >= image.height` branch.
- Backed by `Resizer::classifyAspect()` (static, public — also usable outside the Twig filter for callers needing the bucket without resizing) and `Resizer::resizerAspect()` (instance, the filter callback that dispatches to the existing `resizer()` pipeline).
- Tolerance is hardcoded at 0.1 but overridable via the `timber_kit_resizer_aspect_tolerance` WP filter — keeps the common path one-line at the call site while leaving an escape hatch for design-sensitive components.
- Missing-metadata / non-numeric / zero-dimension sources fall back to `landscape` so legacy assets (pre-ACF imports, SVG without intrinsic dimensions) keep the kit's historical wide-crop default. Components adopting the new filter don't silently shift their rendering for legacy assets.

## Usage

```twig
{{ component_picture({
    image: item.image|resizer_aspect({
        landscape: [['960', '720', '1280', 'crop'], ['480', '360', '', 'crop']],
        portrait:  [['720', '960', '1280', 'crop'], ['360', '480', '', 'crop']],
        square:    [['800', '800', '1280', 'crop'], ['400', '400', '', 'crop']],
    }),
}) }}
```

When the matched orientation bucket has no tuples the helper falls through to `orientations.landscape`; if that's also empty / absent the source is returned unchanged.

## Test plan
- [x] PHPUnit suite — 538 / 538 pass (24 new tests cover landscape / portrait / square classifications, square-band boundaries inside / outside the ±10 % band, tolerance filter overrides — tighten + loosen, missing-metadata fallbacks, source-element selection — last entry wins, and `resizerAspect()` dispatch decisions: empty orientations passthrough, non-array → empty array, matched-but-empty bucket passthrough).
- [x] PHPStan analyse — clean.
- [ ] Manual smoke test in a downstream consumer (pm-a's article-gallery component will swap its inline `is_landscape` branch for `|resizer_aspect` once this lands).

## Follow-ups (out of scope here, tracked separately)
- Styleguide-side parity in `parisek/styleguide` (styleguide-side sub-issue) — registers a parallel filter on the styleguide's bundled-helpers block so demos can show orientation-aware crops without the WordPress runtime.
- Drupal-side parity (Drupal sub-issue) — venue decision (per-project module vs. shared CMS-agnostic package) deferred until at least one active Drupal project hits the pain point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+- `|resizer_aspect` Twig filter — aspect-aware sibling of `|resizer` that classifies a source image's orientation (`landscape` / `portrait` / `square`, ±10 % tolerance around 1:1) and routes to a per-orientation tuple set. Caller passes one map `{landscape: [...], portrait: [...], square: [...]}` instead of branching on `image.width >= image.height` in Twig. Missing-metadata, non-numeric, or zero-dimension sources fall back to `landscape`. Source-element selection mirrors `|resizer` — the last entry of the image list wins. Backed by `Parisek\TimberKit\Resizer::resizerAspect()` (instance method, the filter callback) and `Resizer::classifyAspect()` (static utility, public for callers needing the bucket without resizing). When the matched orientation bucket has no tuples the helper falls through to `orientations.landscape`; if that's also empty / absent the source is returned unchanged. Implements parisek/timber-kit#16.
+- `timber_kit_resizer_aspect_tolerance` WordPress filter — overrides the 0.1 default tolerance band used by `Resizer::classifyAspect()`. Returning a smaller value (e.g. `0.05`) tightens the square band, returning a larger value (e.g. `0.2`) loosens it.
+
 ## [1.5.0] - 2026-05-15
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -31,7 +31,20 @@ Static methods for formatting ACF data into clean arrays for Twig templates:
 
 ### Resizer
 
-Image resizing via [Spatie/Image](https://github.com/spatie/image). AVIF output, responsive variants with breakpoints, crop positions, and cache management. Used as a Twig filter.
+Image resizing via [Spatie/Image](https://github.com/spatie/image). AVIF output, responsive variants with breakpoints, crop positions, and cache management. Exposed as two Twig filters:
+
+- `|resizer( ...$variants )` — caller passes the variant tuples directly.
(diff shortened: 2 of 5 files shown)


parisek:
@copilot review this PR — orientation-aware aspect-bucket dispatch on top of the existing `|resizer` pipeline. Particularly interested in feedback on the missing-metadata fallback chain (landscape default), the floating-point tolerance band, and whether the public `Resizer::classifyAspect()` static surface is named correctly for callers needing the bucket without resizing.

parisek:
Addressed both Copilot findings in 110cea1:

**Finding 1 — `resizerAspect()` doesn't fall back when matched bucket is empty (`src/Resizer.php:758`).** Confirmed bug. `??` only fires on missing keys, so `portrait => []` was treated as "set, but empty" and short-circuited to passthrough instead of using the populated `landscape` tuples — contradicting the CHANGELOG. Tuple selection now does an `is_array() && ! empty()` check on the matched bucket before deciding whether to fall through, matching the documented "empty list = unfilled bucket" semantic.

**Finding 2 — Missing regression test for matched-empty + landscape-populated.** Added `test_resizer_aspect_falls_back_to_landscape_when_matched_bucket_is_empty_array`. To keep the assertion in pure dispatch territory (without touching the filesystem the way real `resizer()` would), the test uses an anonymous `Resizer` subclass that spies on `resizer()`'s `$variants` arg and confirms `resizerAspect()` invoked it with the landscape tuples.

Full suite: 539/539 ✓ · PHPStan: clean ✓.

Resolving this conversation.

parisek:
Addressed the two remaining Copilot threads in 20355c2:

**Boundary IEEE-754 issue (`src/Resizer.php` aspect comparison).** Confirmed. Switched from division-based `abs($width / $height - 1.0) <= $tolerance` to dimension-space `abs($width - $height) <= $tolerance * $height`. Algebraically equivalent for positive dimensions, but IEEE-754-stable at the inclusive boundary — `|1100 - 1000| = 100 <= 0.1 * 1000 = 100` is exact. No magic epsilon, no behavioural change anywhere else (verified against all existing fixtures: 1920×1080, 1110×1000, 1050×1000, 1000×1120, 1099×1000, …).

**Boundary contract not actually verified (`tests/Unit/Resizer/ClassifyAspectTest.php`).** Restored the exact 1100×1000 fixture in `test_classifies_upper_boundary_as_square` — was previously 1099×1000 working around the float bug. The ±10 % inclusive-boundary contract is now locked in by the regression test.

Full suite: 539/539 ✓ · PHPStan: clean ✓. All four Copilot findings (2 suppressed-confidence, 2 inline threads) resolved.

PR #26, opened 2026-05-26
fix(Helpers): collapse 'option'/'options' alias in decodeOptionsNamespace

parisek:
## From which project

Discovered during the **neoli** WordPress theme migration to `parisek/timber-kit:^1.7` — see [portadesign/neoli#17](https://github.com/portadesign/neoli/pull/17), commit [`90530747`](https://github.com/portadesign/neoli/pull/17/commits/90530747).

> **Note from owner:** This pattern has hit multiple migrations now. Each affected project worked around it locally by switching `Helpers::formatFields('option')` → `Helpers::formatFields('options')` (plural). Fixing it upstream means future migrations don't repeat the workaround.

## Why

`Helpers::formatFields('option')` (singular) silently returns `[]` when the registered options page uses ACF's default `post_id` of `'options'` (the default for any `acf_add_options_page()` call without an explicit `post_id`).

### Root cause

`decodeOptionsNamespace()` compares the caller's `'option'` (singular) against the page's `'options'` (plural) via `acf_decode_post_id()`. The helper's docblock claims `acf_decode_post_id()` collapses both forms to a single canonical id:

> Both sides are normalized through `acf_decode_post_id()`, which collapses the `option` / `options` alias to a single canonical id.

But on **ACF Pro 6.x** the alias is **NOT** collapsed:

```php
acf_decode_post_id('option');   // => ['type' => 'option', 'id' => 'option']
acf_decode_post_id('options');  // => ['type' => 'option', 'id' => 'options']
```

`decodeOptionsNamespace` returns the `id` field verbatim, so the caller's `'option'` namespace never matches the page's `'options'` namespace, `getFieldObjectsForOptions()` falls through, and `formatFields('option')` returns `[]`.

### Why the existing test didn't catch it

`FormatFieldsTest::test_options_post_id_resolves_via_acf_get_options_pages` stubs `acf_decode_post_id` to **normalize** both inputs to `id => 'option'`:

```php
Functions\when( 'acf_decode_post_id' )->alias( function ( $id ) {
    return in_array( $id, [ 'option', 'options' ], true )
        ? [ 'type' => 'option', 'id' => 'option' ]
        : [ 'type' => 'post', 'id' => $id ];
} );
```

This fixture papers over the real-world gap. The test passes; production fails.

### Real-world impact

Every project that migrates from a local `classes/Helpers.php` (which accepted `'option'` directly via `get_field_objects()`) to timber-kit's `Helpers` silently loses the entire options-page data unless the caller is also updated to `'options'` (plural). In neoli, this manifested as:

- Empty footer (`footer.email`, `footer.company_info`, etc. all null in `timber_context()`)
- Missing CTA button (`links.header_button`)
- Disabled announcement bar (`announcement.active`)
- Disabled modal (`modal.enabled`)

The data was present in `wp_options` the entire time — the namespace match just kept silently dropping it.

## What it means

`decodeOptionsNamespace()` now canonicalizes the alias internally — both `'option'` (singular) and `'options'` (plural) resolve to `'options'`, matching ACF's default `post_id` value that `acf_get_options_pages()` reports verbatim.

### Diff preview

```diff
 private static function decodeOptionsNamespace( $post_id ): ?string {
     $decoded = acf_decode_post_id( (string) $post_id );
     if ( ! is_array( $decoded ) || ( $decoded['type'] ?? '' ) !== 'option' ) {
         return null;
     }
     $id = $decoded['id'] ?? null;
-    return is_string( $id ) && $id !== '' ? $id : null;
+    if ( ! is_string( $id ) || $id === '' ) {
+        return null;
+    }
+    // Collapse the 'option' / 'options' alias — both refer to the default
+    // options namespace. ACF Pro's acf_decode_post_id() keeps them
+    // separate; canonicalize to the plural form (the default `post_id`
+    // that ACF uses for an options page registered without an explicit
+    // `post_id`, so this matches what `acf_get_options_pages()` reports).
+    if ( $id === 'option' ) {
+        $id = 'options';
+    }
+    return $id;
 }
```

### Documentation impact

The docblock previously promised collapse-via-`acf_decode_post_id()` and was inaccurate. It's been updated to describe what the helper actually does and why ACF Pro 6.x doesn't collapse on its own.

### New regression test

Added `test_options_singular_alias_matches_plural_post_id` to `FormatFieldsTest`. It stubs `acf_decode_post_id` with REAL ACF Pro 6.x semantics (the stub does NOT normalize the alias) and asserts that `formatFields('option')` still finds an options page registered with the default `post_id => 'options'`. Without the fix, this test fails with `assertArrayHasKey('footer', $result)`.

### Sibling projects impacted

Any downstream project that:

- Calls `Helpers::formatFields('option')` (singular — the historical convention from the pre-timber-kit Helpers shim used in `wordpress-base` doctrine and `AGENTS.md`)
- AND registers its options page via `acf_add_options_page()` without overriding `post_id`

…silently loses options-page data after upgrading to a timber-kit release that includes `decodeOptionsNamespace()`. Both forms now work after this fix.

### Test results

```
$ vendor/bin/phpunit tests/Unit/Helpers/FormatFieldsTest.php
PHPUnit 11.5.55 by Sebastian Bergmann and contributors.

....................................                              36 / 36 (100%)

Time: 00:00.061, Memory: 14.00 MB

OK (36 tests, 51 assertions)
```

## Out of scope (left for separate PRs)

- Downstream sync to `wordpress-base` / `tailwind-base` happens explicitly per AGENTS.md doctrine after this merges.
- `decodeOptionsNamespace`'s caller `getFieldObjectsForOptions` is unchanged — only the namespace canonicalization is affected.
- The other `is_string`-path in `Helpers::formatFields()` (line ~679, where `$post_id = $post`) is unchanged — the alias collapse fires later in `isOptionsPostId` → `getFieldObjectsForOptions` → `decodeOptionsNamespace`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `Helpers::formatFields('option')` no longer silently returns `[]` when the registered options page uses ACF's default `post_id` of `'options'` (the default for any `acf_add_options_page()` call without an explicit `post_id`). The previous `decodeOptionsNamespace()` compared the caller's `'option'` (singular) against the page's `'options'` (plural) via `acf_decode_post_id()`, which on ACF Pro 6.x does NOT collapse the alias — the helper's docblock claimed it did, but the existing test fixture papered over the gap by stubbing `acf_decode_post_id` to normalize both forms. Now the alias is canonicalized inside `decodeOptionsNamespace()`, so calls with either form match the page's namespace and surface all registered fields. Discovered during the neoli WordPress theme migration to timber-kit ([portadesign/neoli#17](https://github.com/portadesign/neoli/pull/17)) — symptom was an empty footer + missing global ACF data despite the data being present in `wp_options`. New regression test (`test_options_singular_alias_matches_plural_post_id`) mirrors real ACF Pro 6.x semantics (no normalization in the stub) so the fix can't silently re-regress.
+
 ## [1.7.0] - 2026-05-25
 
 ### Added
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -836,9 +836,11 @@ class Helpers {
 	 * stores its fields under `company_settings`, not under the default
 	 * `option` namespace — so `formatFields('option')` must not surface them,
 	 * and `formatFields('company_settings')` must not pick up unrelated
-	 * default-namespace pages. Both sides are normalized through
-	 * `acf_decode_post_id()`, which collapses the `option` / `options` alias
-	 * to a single canonical id.
+	 * default-namespace pages. Both sides are routed through
+	 * `decodeOptionsNamespace()`, which canonicalizes the default-namespace
+	 * `option` / `options` alias to a single id (`'options'`). ACF Pro's
+	 * own `acf_decode_post_id()` does NOT collapse the alias on its own —
+	 * see {@see decodeOptionsNamespace()} for details.
 	 *
 	 * Multiple pages sharing the same namespace are still unioned. Fields
 	 * with the same `name` across same-namespace pages collide on
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -895,8 +895,17 @@ class Helpers {
 	 * string doesn't decode to an options namespace.
 	 *
 	 * Used to compare a caller's `$post_id` against each registered options
-	 * page's `post_id` after collapsing the `option` / `options` alias —
-	 * `acf_decode_post_id()` returns the same `id` for both forms.
+	 * page's `post_id` after collapsing the `option` / `options` alias to a
+	 * single canonical id.
+	 *
+	 * ACF Pro's `acf_decode_post_id()` does NOT collapse the alias on its own:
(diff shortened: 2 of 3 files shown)


parisek:
## Codex code review applied

Independent second-opinion review (via Codex CLI rescue) returned **Approve with concerns** on the original commit. All three concerns addressed in 70f3acc:

1. **Stale docblock at `src/Helpers.php:835-841`** — removed the factually-wrong claim that `acf_decode_post_id()` collapses the alias. Rewritten to describe what the code actually does (collapse happens inside `decodeOptionsNamespace()`).

2. **Stale docblock at `tests/.../FormatFieldsTest.php:856-858`** — updated to acknowledge that the older test's loose `acf_decode_post_id` stub (which normalizes both forms) doesn't match real ACF Pro 6.x semantics, but the assertion still holds. Cross-references the new sibling test.

3. **Missing WPML coverage** — added `test_options_wpml_prefixed_namespace_is_preserved`. Regression guard: `formatFields('options_en')` must match `post_id => 'options_en'` page AND must NOT leak the default-namespace `post_id => 'options'` page. Confirms the exact-match `'option' → 'options'` rewrite is appropriately conservative.

Full suite: 37/37 FormatFieldsTest tests pass.

Issue #29, opened 2026-05-28
Proposal: WpmlBlockOverride — runtime source-value override for Copy fields in ACF Gutenberg blocks


parisek:
# WpmlBlockOverride — Proposal & Research Findings

> **Single consolidated reference** for everything discovered during research session on 2026-05-28. Linked from atelier99 working session (branch `feature/wpml-block-render-override`).

## TL;DR

**Problem:** Editor changes an image in an ACF Gutenberg block in the source language. The change never propagates to translation frontends. Forcing sync requires manual "Update translation" via WPML > Translation Management → ATE re-job for **every Copy field change**. This is documented as a recurring WPML pain point since 2019.

**Root cause:** WPML/ACFML triggers `needs_update` only when *translatable* (text) content changes. A change in a Copy field (`wpml_cf_preferences = 1`) — typically image/file — is not a translatable change, so the flow never runs. Even if it did, translation `post_content` is only rewritten during an ATE import job.

**Proposed solution:** A `final class WpmlBlockOverride` in `parisek/timber-kit` that hooks `render_block_data` at priority 20 (after WPML's own handlers) and, for Copy fields in ACF blocks rendered in a non-default language, overrides `attrs.data.<field>` with the value from the source-language post — at render time, no DB writes.

**Result:** ACF configuration becomes the single source of truth for Copy fields. Zero drift. Zero sync mechanism. Zero admin UI. The Translate flow remains entirely handled by ACFML/ATE as today.

---

### 1.1 Block storage

ACF Gutenberg blocks serialize into `post_content` as HTML comments:

```html
<!-- wp:acf/jumbotron-video {"id":"block_5fd76xyz","name":"acf/jumbotron-video","data":{"title1":"…","background_image":456,"video":789},"mode":"preview"} /-->
```

Key observations:
- **Every language version of the post has its own `post_content`** — each with its own serialized block instances and values.
- The `data` object holds ACF field values (scalars: text, ID, URL).
- The `id` attribute (e.g., `block_5fd76xyz`) is **stable across languages** — generated at source insert, inherited by translations. This is our pairing key for source ↔ translation block instances.

### 1.2 WPML translation flow

When the editor saves a source post:

1. WPML diffs new vs old.
2. If a translatable string changed (e.g., a Translate-field title), translations are marked `needs_update`.
3. With "Translate Everything Automatically" + WPML AI/DeepL, an ATE job is queued automatically.
[...]

PR #45, opened 2026-06-05
feat(wpml): WpmlBlockOverride — render-time Copy-field sync for ACF blocks

parisek:
Implements #29. **Draft** — class + unit tests are done; opening as draft because the two spec items flagged *"to verify during implementation"* (ACFML preference auto-detection + attachment-id remap) need validation against a live **WPML + ACFML** install. Integration test target: **keypers**.

## The problem (#29)

When an editor changes a **Copy** field (typically an image) in the source language, WPML never marks the translation `needs_update` — a Copy change isn't a *translatable* change — so the translation's `post_content` keeps the stale value until someone manually re-runs the ATE job. Documented WPML pain since ACFML 1.3 (2019), unsolved in core.

## The fix

Opt-in: `\Parisek\TimberKit\WpmlBlockOverride::register()`. Hooks `render_block_data` at **priority 20** (after WPML's own handlers) and, for ACF blocks rendered in a non-default language, overwrites each Copy field's `attrs.data.<field>` with the value from the **source-language** post — at render time, **no DB writes**. ACF config becomes the single source of truth for Copy fields; the Translate flow stays entirely with ACFML/ATE.

## Data flow (per spec §4.3)

`shouldOverride()` (bypass non-acf/admin/REST/source-lang) → `getCopyFields()` (ACFML "Copy" preference, `_`-skip, 24h transient, + `timber_kit/wpml_block_override/copy_fields` filter) → `getSourcePostId()` (`wpml_object_id`) → `getSourceBlocks()` (`parse_blocks`, per-request memo) → `findSourceBlock()` (match by stable `attrs.id`, recurses `innerBlocks`) → `applyCopyFields()` (overwrite + `_field`-key companion + attachment remap) → safe-degrade on drift/missing id.

Single `final class`, static methods, no Timber dependency, `timber_kit/<feature>/<event>` filter naming — matches the kit's `BlockRenderer` pattern.

## Two extension/validation points

- **Copy-field detection** auto-reads the ACFML preference, but a project can declare its Copy fields explicitly through the `timber_kit/wpml_block_override/copy_fields` filter — so keypers can wire it immediately while auto-detection is confirmed.
- **Attachment remap** runs numeric Copy values through `wpml_object_id` for `attachment` (current language's duplicated media, source id fallback).

These two are exactly what the keypers integration test validates.

## Tests / checks

- **24 unit tests** (Brain Monkey): `FindSourceBlock` (id match + innerBlocks recursion + null safe-degrade), `ApplyCopyFields` (overwrite, `_field` companion, non-Copy untouched, attachment remap, non-numeric passthrough), `ShouldOverride` (acf/admin/source-lang/missing-lang), `GetCopyFields` (Copy-preference collection, `_`-skip, explicit filter, transient cache), `Register` (priority 20 + cache flush), `Filter` (non-array, source-lang bypass, full end-to-end override).
- **Full suite: 713 green.**
- **PHPStan level 8: clean.**

## Out of MVP (per spec §5)

Nested Repeater/Group/Flexible Copy fields; REST API output (server-rendered project); WPML "Different fields" mode (untested — to verify).


parisek:
Closing as a duplicate of #31.

Both PRs implement the same feature (WpmlBlockOverride — render-time Copy-field sync for ACF blocks). This branch (`feat/wpml-block-override`) was cut fresh from main a week after #31 and reimplemented the feature from scratch in a single commit, independently of #31's history — most likely without awareness that #31 already existed and had been through two review passes.

#31 is the implementation we keep. It is more complete and correct on the two hardest points:
- **Block pairing**: #31 matches source↔translation blocks by position (guarded by a structural-integrity count check). This PR matched by `attrs.id`, which real ACF blocks do not carry in `post_content` (the `block_*` id is generated at render by `acf_get_block_id()`, not serialized) — so on real content findSourceBlock would never match and the override would be a silent no-op. The green tests here inject a synthetic `attrs.id`.
- **Nested fields**: #31 walks repeater/group sub_fields and rebuilds the flattened keys (`steps_0_image` …). This PR only handled top-level `attrs.data.<field>`.

#31 also remaps post/term references (not just attachments) and ships README docs.

Kept here for the record so the duplication is traceable. Continuing on #31.

PR #51, opened 2026-06-15
feat(media): $enabled_image_sizes — opt-in allowlist for intermediate sub-sizes

parisek:
## Summary

- Adds `StarterBase::$enabled_image_sizes` (`string[]|null`, default `null`) — opt-in allowlist of image sub-size slugs generated on upload
- Hooked to `intermediate_image_sizes_advanced`, which runs inside `wp_create_image_subsizes()` for every upload path (media library, REST, WP-CLI sideload)
- When `null` (default): no hook registered, WP behaviour unchanged
- When set: only listed slugs survive; everything else (including WP 5.3+ `1536x1536` and `2048x2048`) is dropped for new uploads

## Motivation

Projects typically use 3–4 image sizes in Twig templates, but WordPress generates 6+ by default. The two 5.3+ sizes (`1536x1536`, `2048x2048`) add significant disk usage and upload time without being referenced by any template. This property gives projects surgical control with zero overhead when not used.

## Example

```php
// In Base::__construct(), before parent::__construct():
$this->enabled_image_sizes = ['thumbnail', 'medium', 'large'];
// → 1536x1536 and 2048x2048 are no longer generated on new uploads
```

## Scope

- New uploads only — existing images are unchanged. `wp media regenerate` + the same `$enabled_image_sizes` setting applies the allowlist to existing uploads.
- Does not touch `big_image_size_threshold` (max dimension cap) — orthogonal concern.
- Does not auto-disable any size — opt-in only, default is WP default.

## Test plan

- [ ] `FilterImageSizesTest` — 7 unit tests: passthrough when null, allowlist filtering, empty allowlist, unknown slug ignored, settings preserved unmodified, meta/id unused
- [ ] `RegisterMediaHooksTest` — 2 new tests: hook registered when allowlist set, hook skipped when null
- [ ] Full suite: `composer test` → 790 tests, 0 failures
- [ ] `composer phpstan` → no errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


parisek:
Closing — not wanted. The $enabled_image_sizes feature was implemented outside its intended scope; revisit separately if/when needed.

PR #58, opened 2026-06-21
feat: $minify_style flag for non-minified builds

parisek:
## What

Adds a `$minify_style` boolean property (default `true`) to `StarterBase`. When `false`, `assets()` always enqueues `style.css` instead of `style.min.css` in production.

## Why

Projects like `atelier99` whose build pipeline emits only `style.css` (no `.min`) previously had to override `assets()` entirely just to change one filename. With this flag they can set `$minify_style = false` and use the native `assets()` as-is.

## Behaviour

- Default `true` → `style.min.css` in production, `style.css` under `WP_DEBUG` (no change from today).
- `false` → always `style.css`, `WP_DEBUG` still forces `style.css` regardless.

## Tests

Added `tests/Unit/StarterBase/AssetsTest.php` with two cases: default enqueues `.min.css`, and `$minify_style = false` enqueues `style.css` (not `.min.css`). All 789 tests pass, 0 errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


parisek:
Closing — decided not to add $minify_style to the library. atelier99 keeps its assets() override for the non-minified build instead.

PR #90, opened 2026-07-20
feat(wpml): keep theme domain authoritative (exclude from String Translation)

parisek:
## From-Project

fellows (WordPress + WPML) — a theme string was corrected in the `.po` file, the `.mo` was rebuilt, and the site kept rendering the old value. Root cause: WPML String Translation had already scanned the theme's `.mo`, registered the strings into ST, and compiled its own overriding `wp-content/languages/wpml/<domain>-<locale>.mo` — which WPML's Just-In-Time MO loader serves *instead of* the theme's own `.mo` at runtime. Editing the theme's `.po` source of truth silently had no effect.

## Why

Verified against WPML String Translation 3.5.x source: a text-domain listed in `icl_sitepress_settings['st']['wpml_st_auto_reg_excluded_contexts']` is (a) excluded from ST auto-registration (`Settings::isDomainRegistrationExcluded()`), and (b) skipped by WPML's Just-In-Time `.mo` loader (`MOFactory::get()` iterates every domain except the excluded ones — `'default'` is always excluded there too). Adding the theme's own domain to that list keeps the theme's `.po`/`.mo` pair — dev-managed, version-controlled — as the single, authoritative source for theme strings, instead of silently forking into a second WPML-managed copy.

## What

New StarterBase flag, `protected bool $wpml_theme_domain_authoritative = true;` (default on, same deliberate exception to the default-off flag doctrine as `$wpml_skip_empty_translation_job_fields` — the off-by-default state is the harmful one here). When on, registers a filter on `option_icl_sitepress_settings` that injects `$this->theme_name` into the excluded-contexts array **at read time only** — no `update_option()` call, nothing persisted to the database, applies everywhere WPML reads the setting. No-ops entirely on non-WPML sites (the option filter simply never fires).

```php
protected bool $wpml_theme_domain_authoritative = true;

// registered in registerMiscHooks(), alongside the sibling WPML flags
if ( $this->wpml_theme_domain_authoritative ) {
    add_filter( 'option_icl_sitepress_settings', array( $this, 'wpml_exclude_theme_domain_from_st' ) );
}

public function wpml_exclude_theme_domain_from_st( $settings ) {
    // ...injects $this->theme_name into $settings['st']['wpml_st_auto_reg_excluded_contexts']
}
```

## Migration note for existing sites — now automated via WP-CLI

Sites upgrading to this version where WPML String Translation has **already** registered the theme's domain need a one-time cleanup, since the flag only prevents *future* registration/override. This is now a single command instead of a manual admin-UI pass:

```
wp timber-kit wpml-cleanup-theme-domain --dry-run
wp timber-kit wpml-cleanup-theme-domain
```

`WpmlCleanupThemeDomainCommand` (new `src/Cli/WpmlCleanupThemeDomainCommand.php`, backed by the unit-tested pure `src/Wpml/ThemeDomainCleanupPlan.php`):

1. Resolves the target domain (`--domain=<name>`, default the theme's own `TextDomain`) and guards on WPML String Translation being active (`icl_strings` table present).
2. Counts, then (JOIN-style, children before parent) deletes `icl_string_translations` and `icl_string_positions` rows keyed off `icl_strings.context = <domain>`, then the `icl_strings` rows themselves.
3. Globs and deletes the compiled `wp-content/languages/wpml/<domain>-*.{mo,l10n.php,json}` files, and fires `wpml_cache_clear`.
4. `--dry-run` reports counts/files without deleting anything; a real delete prompts for confirmation unless `--yes` is passed.

Verified against a real site (fellows) before this PR: 102 `icl_strings`, 69 `icl_string_translations`, 27 `icl_string_positions` rows for the theme's domain — all safely removable once the domain was excluded, since WPML's JIT MOFactory already skips the excluded domain and the theme's own `.po`/`.mo` is authoritative.

Sites that haven't hit this yet (fresh installs, or WPML ST never scanned the theme) need no manual step — the flag is preventative by default, and the cleanup command reports "nothing to clean up" for them.

## Test plan

- [x] `php -l` on all changed/added files — no syntax errors
- [x] `composer test` (PHPUnit Unit suite, incl. new `tests/Unit/Wpml/ThemeDomainCleanupPlanTest.php`) — green
- [x] `phpstan analyse` (level 8) — no errors
- [ ] CI (this repo has GitHub Actions — waiting on checks before calling this ready)

https://claude.ai/code/session_01Fgb38FaYZjZufuA7CYyowq

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- `$wpml_theme_domain_authoritative` StarterBase flag (**default on**, same deliberate exception as `$wpml_skip_empty_translation_job_fields`): keeps the theme's gettext text-domain authoritative over WPML String Translation by runtime-injecting it into `icl_sitepress_settings['st']['wpml_st_auto_reg_excluded_contexts']` via a filter on `option_icl_sitepress_settings` — no `update_option()` write, nothing persisted. Without it, WPML ST scans the theme's compiled `.mo`, registers every string, and compiles its own overriding `wp-content/languages/wpml/<domain>-<locale>.mo` that WPML's Just-In-Time MO loader serves instead of the theme's own `.mo` — so editing the theme's `.po` and rebuilding has no visible effect, the corrected string keeps rendering its stale WPML-compiled value (hit in production on fellows, 2026-07-20). No-ops without WPML. Opt out with `false` in the project's `Base` if a project wants translators managing theme strings through WPML ST instead of the `.po`/`.mo` pipeline.
+
 ## [1.25.0] - 2026-07-16
 
 ### Added
--- a/src/Cli/WpmlCleanupThemeDomainCommand.php
+++ b/src/Cli/WpmlCleanupThemeDomainCommand.php
@@ -1,0 +1,237 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Parisek\TimberKit\Cli;
+
+use Parisek\TimberKit\Wpml\ThemeDomainCleanupPlan;
+
+/**
+ * `wp timber-kit wpml-cleanup-theme-domain` — purge leftover WPML String
+ * Translation rows and compiled translation files for a text domain that
+ * `$wpml_theme_domain_authoritative` has excluded from ST.
+ *
+ * Once a project runs with `$wpml_theme_domain_authoritative` ON (the
+ * default — see
+ * {@see \Parisek\TimberKit\StarterBase::wpml_exclude_theme_domain_from_st()}),
+ * WPML stops registering new strings for the theme's domain and stops
+ * compiling an overriding `.mo` for it. Rows String Translation had already
+ * registered *before* the exclusion took effect are left behind: inert
+ * (WPML's Just-In-Time MO loader skips the excluded domain, and ST no
+ * longer surfaces them to translators) but still occupying
+ * `icl_strings` / `icl_string_translations` / `icl_string_positions`, plus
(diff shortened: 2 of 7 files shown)


parisek:
Two open questions from Codex review that need an owner call before this is ready for review (implementation-side fixes from the same review — empty-domain guard + edge-case tests — are already pushed):

**1. Default value.** `$wpml_theme_domain_authoritative` defaults to `true`. For any downstream site that currently relies on WPML String Translation to manage the theme's domain strings, upgrading `parisek/timber-kit` would silently stop ST from managing them (existing ST-registered strings still show in the admin, but new/changed theme strings stop flowing to ST). Two options:
   - Keep default `true`, with a prominent CHANGELOG/migration note (as already drafted) — matches this repo's stated exception to the default-off flag doctrine, same rationale as `$wpml_skip_empty_translation_job_fields`.
   - Default `false` (opt-in) — safer for existing installs, but then every project hitting this desync independently has to discover and opt in themselves (the problem the flag exists to prevent).

**2. Filter timing — needs verification on a live WPML site.** The fix works by filtering `option_icl_sitepress_settings` at read time (non-persisting, nothing written to the DB). This is correct *if* WPML calls `get_option( 'icl_sitepress_settings' )` fresh on every relevant read, after this filter is registered, and doesn't cache an earlier in-memory copy from before the filter attached. The fix for the fellows production symptom was verified via a **persisted** option value (the actual bug + fix), not via this filter-only mechanism — the runtime-filter path itself is unverified against a live WPML site. Worth confirming before relying on it broadly. If the filter proves unreliable (WPML reads the option once early and caches it), a proven fallback is persisting the setting once via `update_option()` on `admin_init` (guarded to run only when the value is actually missing, to stay idempotent and avoid an unconditional write per request).

PR #93, opened 2026-07-29
fix(helpers): let a context build defer a form embed instead of rendering it

### LINKED ISSUE #92: formatFields('option') renders heavy embeds on every request
## Problem

`Helpers::formatFields( 'option' )` eagerly formats **every** field on the options page. When one of them is a `post_object` pointing at a form (`wpforms`, `contact-form-7`), `fieldFormatter()` runs it through `do_shortcode()` and returns fully rendered HTML.

`timber-kit.md` § *Override `timber_context()`* documents `Helpers::formatFields('option')` as **the canonical way** to source global options. So the render happens on **every single request**, on every page — including pages that contain no form at all.

Measured on a production site (oekoplan, timber-kit v1.26.0):

```
$x = Helpers::formatFields("option");
gettype($x["career"]["form"])  →  string
strlen(...)                    →  6050
substr(..., 0, 60)             →  <div class="wpforms-container wpforms-render-modern" id="wpforms-1991…
```

Two consequences:

1. **6 kB of form HTML is rendered per request** just to build the global context, then thrown away on every page that doesn't use it.
2. **WPForms registers its CSS/JS globally.** Rendering the shortcode triggers the plugin's asset enqueue, so `wpforms-*.css`, the layout stylesheets and jQuery land in `<head>` of every page on the site. Verified by diffing rendered HTML before/after the theme adopted StarterBase — 6 stylesheets + jQuery appeared on pages with no form.

There is also a correctness angle: a form rendered into the global context means duplicate `id="wpforms-form-…"` in the DOM if the same form is rendered again by the page itself.

## Where it comes from

`src/Helpers.php` around line 1216:

```php
} elseif ( $field['value']->post_type === 'wpforms' ) {
    if ( $is_preview ) {
        $field['value'] = '[wpforms id="' . $field['value']->ID . '"]';
    } else {
        $field['value'] = do_shortcode( '[wpforms id="' . $field['value']->ID . '"]' );
    }
}
```

The `field_formatter_*` filter at line ~1303 fires **after** this, so a downstream filter can replace the output but cannot prevent the render or the asset enqueue.

## Current workaround (and why it isn't the fix)

Passing the existing second parameter:

```php
$global_fields = Helpers::formatFields( 'option', true );  // $is_preview
```

This works — verified that nothing else changes:

```
links + footer + announcement with preview=true vs without  →  byte-identical
career.form with preview=true                               →  [wpforms id="1991"]
```

But `$is_preview` means *"we are in the editor preview"*. Using it to mean *"don't render heavy embeds"* overloads a flag with unrelated semantics, and any future behaviour keyed on `$is_preview` would silently start applying to normal front-end requests.

## Proposed solutions

Roughly in order of preference:

1. **Lazy `post_object` formatting.** Return a small value object that renders the shortcode on string conversion. Twig calling `{{ content.form }}` renders it; a context build that never touches it pays nothing. No API break, fixes every call site at once.
2. **A dedicated parameter/flag** — `formatFields( $post, $is_preview, $render_embeds = true )`, or an options array. Explicit, but every caller has to know to pass it.
3. **A filter that fires _before_ the render**, e.g. `field_formatter_pre_post_object`, so a theme can opt out per field. Smallest change, but leaves the default behaviour surprising.

Happy to open a PR for whichever direction you prefer — (1) reads as the real fix to me.

## Context

Found while migrating a theme to StarterBase. Reported from oekoplan (`~/Sites/wordpress/oekoplan`), timber-kit v1.26.0, ACF Pro 6.8.6, WPForms 2.0.0.2.

parisek:
Closes #92.

> **Rebuilt 2026-08-27 on current `main`.** The original branch was one commit from before v1.26 and conflicted in all three files it touched. The approach changed too, so it was re-implemented rather than merged forward.

## The problem

`Helpers::formatFields( 'option' )` is the documented way to source globals from `timber_context()`. It formats **every** field on the options page whether or not the page will print it, and a `post_object` pointing at a form is formatted by running its shortcode. So every request of every page rendered a form — 6 kB of markup on the reported site, built and discarded.

## The original report named the wrong mechanism for the asset cost

Worth correcting, because it decides the shape of the fix. The report said the render "triggers the plugin's asset enqueue, so `wpforms-*.css` and jQuery land in `<head>` of every page".

`Frontend.php:1547` says otherwise — the `wp_head` path enqueues CSS only when `is_singular()` **and** the post content holds the shortcode or block:

```php
if ( ! is_singular() ) { return; }
if ( has_shortcode( $post->post_content, 'wpforms' ) || has_block( 'wpforms/form-selector' ) ) {
    $this->assets_css();
}
```

On a form-less page neither is true, so that path never fired. What loads the assets is `assets_footer()` on `wp_footer`, which enqueues whenever `$this->forms` is non-empty — and the context build populated it by rendering.

Same symptom, different hook. It matters because it means suppressing the *output* is not enough; the fix has to stop the *render*, since that is what registers the form.

## The fix

`formatFields( $post, $is_preview, $render_embeds = true )` and the matching `fieldFormatter()` parameter. `false` hands back `[wpforms id="12"]` instead of rendering it. Default unchanged — no existing caller moves.

**Plus the half the earlier draft lacked:** a `|shortcode` Twig filter, so the one template that actually prints the form renders it where it is printed.

```twig
{{ content.contact.form|shortcode }}
```

Rendering there is when the plugin discovers the form. WPForms handles that case explicitly — `assets_footer()` exists for it, at the cost of a brief flash of unstyled content, which is the trade the deferral buys on every page that has no form at all. **Verified in the plugin source rather than assumed**; it is the question that decides whether deferring is viable at all, and nobody had checked it.

## Rejected: the lazy value object

The issue ranked this first ("reads as the real fix to me"). Three concrete objections:

- **It is never empty.** `formatFields()` drops fields whose formatted value is empty. A lazy object is always truthy, so an unfilled form field would start appearing in the context.
- **It is not storable.** `MenuFieldsCache::isStorable()` rejects objects, so an options page holding one could never be cached — the opposite of what deferral is for.
- **It breaks `is_string()`.** A theme testing the value's type gets a different answer, silently.

## It composes with the menu cache

A rendered form carries a nonce, so `fieldFormatter()` counts it dynamic and nothing may cache it. **Deferring the embed is what lets the rest of an options page be storable at all** — which is what #147 needs, and it follows the same rule as the menu cache ([ADR-0007](../blob/main/docs/adr/0007-prove-cache-purity-from-inputs.md)): decide from an input, before the work.

`test_deferring_the_embed_keeps_the_value_cacheable` asserts exactly that, by reading the dynamic counter either side of both calls.

## Migration for a project that hits this

Two steps, and the second only where a form is actually printed:

```php
// timber_context()
$global_fields = Helpers::formatFields( 'option', false, false );
```
```twig
{# the one template that prints it #}
{{ content.contact.form|shortcode }}
```

## Tests

Four added, verified load-bearing by removal — disabling the deferral fails exactly three of them and nothing else. 1828 tests, PHPStan clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- `Helpers::formatFields( $post, $is_preview, $render_embeds = true )` and the
+  matching `fieldFormatter()` parameter. `false` hands back the shortcode for a
+  `post_object` field pointing at a form (`wpforms`, `wpcf7_contact_form`)
+  instead of rendering it. The default is unchanged, so no existing caller
+  moves.
+
+  Pass it when building a context. A context build formats every field on the
+  options page whether or not the page will print it, and the documented
+  canonical call — `Helpers::formatFields( 'option' )` from `timber_context()`
+  — therefore rendered a form on **every** request of **every** page. Measured
+  on a production site: 6 kB of form markup built and discarded per request.
+
+  The asset cost is the larger half and the original report named the wrong
+  mechanism for it. WPForms enqueues in `wp_head` only when `is_singular()` and
+  the post content holds the shortcode or block; what loads its stylesheets and
+  jQuery site-wide is the `wp_footer` pass, which fires because a rendered form
+  registered itself during the context build. Same outcome, different hook, and
+  it matters: the fix has to stop the render, not the output.
+
+  A second copy of the form's element ids also landed in the DOM of any page
+  that rendered the same form itself.
+
+- `|shortcode` Twig filter, the other half. A deferred embed is a string, so the
+  one template that prints the form renders it there:
+  `{{ content.contact.form|shortcode }}`. Rendering at that point is when the
+  form plugin discovers the form; WPForms handles it explicitly by enqueueing on
+  `wp_footer`, at the cost of a brief flash of unstyled content — which is the
+  trade the deferral buys on every page that has no form at all.
+
+  A rendered form carries a nonce, so `fieldFormatter()` counts it as dynamic
+  and nothing may cache it. Deferring the embed is therefore also what lets the
+  rest of an options page be storable — the same rule the menu cache follows
+  ([ADR-0007](docs/adr/0007-prove-cache-purity-from-inputs.md)).
+
+
+### Added
+
 - `CacheSignature` — the shared part of every cross-request cache key: site,
   language, the current user's roles, and a content version from
   `wp_cache_get_last_changed()` over `posts` and `terms`. WordPress bumps those
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -794,6 +794,25 @@ class Helpers {
 	 * all, and reads identically to "the data is not there". `get_field()` by
 	 * name still resolves in that situation; prefer it when probing.
 	 *
+	 * **Pass `$render_embeds = false` when building a context.** A context build
+	 * formats every field on the options page whether or not the page will print
+	 * it, and a `post_object` pointing at a form is rendered by running its
+	 * shortcode. That spends the render on every page — measured at 6 kB of
+	 * markup on a site with one such field — and it is what makes the form
+	 * plugin load its stylesheets and jQuery site-wide: a rendered form
+	 * registers itself, and the plugin's `wp_footer` pass then enqueues assets
+	 * for it. It also puts a second copy of the form's element ids into the DOM
+	 * of any page that renders the same form itself.
+	 *
+	 * The template that actually prints the form renders it there:
+	 * `{{ content.contact.form|shortcode }}`.
+	 *
+	 * There is a second effect worth knowing. A rendered form carries a nonce,
+	 * so it is never a function of stored data and nothing may cache it —
+	 * `fieldFormatter()` counts it as dynamic for exactly that reason.
+	 * Deferring the embed is therefore also what lets the rest of an options
+	 * page be storable at all.
+	 *
 	 * @param object|int|string|null $post       Post object, term object, numeric post ID,
 	 *                                            options-page string key, or null to use the
 	 *                                            current queried object.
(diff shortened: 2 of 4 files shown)


parisek:
## Zůstává draft — napřed by to chtělo změřit reálný přínos

Nález sám o sobě platí a měření z issue je pravdivé: `formatFields('option')` opravdu vyrenderuje 6 kB formuláře na každém requestu. Otázka je, jestli to v praxi za tuhle změnu API stojí — a zatím pro to nemáme čísla.

Dva důvody, proč jsem skeptický:

**1. Formulář na options page má menšina projektů.** Není to vzor, na kterém by flotila stála — je to konfigurace, která se hodila jednomu webu (`career.form` na kariérní stránce). Optimalizace, která se týká pár projektů, ale rozšiřuje veřejnou signaturu `formatFields()` a `fieldFormatter()` pro všechny, má nepříznivý poměr.

**2. I tam, kde ten formulář je, se stejně obvykle načítá globálně.** Když má web kontaktní formulář v patičce, WPForms si svoje CSS/JS zaregistruje na skoro každé stránce tak jako tak. Ušetřený render je pak 6 kB HTML, ne ta assets stopa, kvůli které to celé vypadalo vážně. Právě ta stopa přitom byla hlavní argument.

Čili než tohle mergnout, chtělo by to změřit na konkrétním webu:

- kolik stránek reálně načte WPForms assets **s** opt-outem a kolik **bez** něj
- jestli po opt-outu vůbec nějaká stránka assets ztratí, nebo je patička načte znovu
- jaký je rozdíl v TTFB / počtu dotazů, ne jen v bajtech vyrenderovaného HTML

Kdyby se ukázalo, že úspora je jen těch 6 kB zahozeného HTML a assets se stejně načtou z patičky, je správná odpověď **neměnit API** a nechat to být.

### Co s tím zatím

PR nechávám jako draft. Kód je hotový a otestovaný (5/5 CI zelených, dvě nezávislá code review bez nálezu), takže když se měření vyplatí, dá se vzít tak, jak je.

Mezitím dvě věci, které z toho zůstávají užitečné bez ohledu na výsledek měření:

- **Pojmenování.** `$render_heavy_embeds` slibuje víc, než umí — gatuje jen dvě ze čtyř `do_shortcode()` větví ve `fieldFormatter()`. `wysiwyg` (ř. 1159) a `textarea` (ř. 1166) pouštějí shortcody dál, takže `[wpforms id="…"]` napsaný editorem do wysiwyg pole na options page projde i s opt-outem. Kdyby se PR někdy bral, název by měl být užší (`$render_form_objects`), nebo by se musel rozšířit dopad — a to druhé má vlastní riziko (syrový `[year]` vysázený do stránky).
- **Ta wysiwyg/textarea mezera** platí nezávisle na tomhle PR. Jestli se ukáže, že problém stojí za řešení, je to spíš argument pro filtr před renderem než pro parametr — pokryl by všechny čtyři cesty jedním mechanismem.

PR #99, opened 2026-08-01
fix(helpers): gate the repeater/flexible null-value pass-through on $is_preview

### LINKED ISSUE #98: fieldFormatter(): gate the repeater/flexible null-value pass-through on $is_preview
## Summary

`Helpers::fieldFormatter()` returns the **raw ACF field definition** for an unfilled `repeater` / `flexible_content` regardless of context. That is deliberate — the null-value pass-through preserved by #68 for the block-preview contract — but it is currently ungated, so it also fires on ordinary front-end reads, where the definition array masquerades as a populated list.

This is not a request to revert #68. It is a request to gate the carve-out on the `$is_preview` argument that `fieldFormatter()` already accepts.

## Where it bites

`formatMenu()` merges per-item ACF fields into every menu item:

```php
$acf_fields = (array) Helpers::formatFields( $item );   // Helpers.php:1546 — $is_preview defaults to false
```

A `nav_menu_item` group with a `more_items` repeater, on a site where only 10 of 425 items ever filled it:

```
more_items PHP type histogram:   [array] => 69
more_items count histogram:      [28] => 67,  [1] => 1,  [2] => 1
first entry keys:                ID, key, label, name, aria-label, prefix, type, value
```

67 of 69 menu items report a 28-element `more_items`. The 28 elements are the field definition's own keys.

The failure is quiet, which is what makes it worth reporting. A template guard of the documented shape passes:

```twig
{% if item.more_items and item.more_items|length > 0 %}   {# true: length is 28 #}
  <ul class="border-t">
    {% for more_item in item.more_items %}
      {% if more_item.title %}                            {# false for all 28 keys #}
```

No junk rows appear, so nothing looks broken — but the wrapper has already opened. The visible result is an empty `<ul>` whose `border-t` draws a stray rule across the panel. Measured on one page: 3 rails rendered, 1 of them empty. It survived review precisely because the inner guard masked the outer one.

`icon` (image) and `integrations` (group) on the same items behave correctly; only the repeater degrades this way, which matches the carve-out exactly.

## Why gating on `$is_preview` is the right shape

The signature already carries the discriminator, and the callers already set it correctly:

| Call site | `$is_preview` | Wants the definition? |
| --- | --- | --- |
| `BlockRenderer::buildContent()` → `formatFields( $post_id, $is_preview )` (BlockRenderer.php:295) | propagated | **yes** — unsaved block needs `sub_fields` to render a placeholder |
| `formatMenu()` → `formatFields( $item )` (Helpers.php:1546) | `false` | no — wants data |
| theme/page reads → `formatFields( $post )` | `false` | no — wants data |

So the proposal is one condition:

```php
if ( ! isset( $field['value'] ) ) {
    if ( $is_preview && in_array( $field['type'], array( 'repeater', 'flexible_content' ), true ) ) {
        return $field;
    }
    return FALSE;
}
```

`$is_preview` is currently consulted only for the CF7 / WPForms shortcode branches, so this extends an existing convention rather than inventing one.

## Is it a breaking change?

**Public API: no.** No signature, no return type, no new argument.

**Behaviour: yes, narrowly — and it must be declared.** Exactly one input changes: `repeater` / `flexible_content` with an explicit `value => null`, read with `$is_preview = false`. Before: field-definition array. After: `FALSE`, which `mapFields()` drops, so the key is simply absent.

Two things make me call it minor rather than patch:

1. **Two existing tests encode the current behaviour** — `test_repeater_null_value_returns_field_as_is` and `test_flexible_content_null_value_returns_field_as_is`, both named in #68. They call `fieldFormatter()` without `$is_preview`, so they would need to pass `true` to keep asserting the preview contract. A change that edits tests asserting a documented contract is a contract change, whatever the diff size.

2. **There is one way to regress a consumer**: a block render path that does *not* go through `BlockRenderer` and forgets to pass `$is_preview = true`. It would lose its editor placeholder — visible in the editor, not on the front end, and fixed by passing the flag the parameter already exists for.

I could not find a legitimate consumer of the current behaviour: outside preview it hands a definition array to code expecting a list, and #68's own rationale — *"must read as empty, not leak the raw field-definition array into templates"* — is the argument for closing this last case, not against it.

**Migration note if accepted:** consumers that render ACF blocks outside `BlockRenderer` should confirm they pass `$is_preview`. Nothing else needs to change; front-end reads only stop receiving a value they could not have used.

## Alternative considered and rejected

Returning `array()` instead of `FALSE` for the non-preview case would keep `is iterable` guards working and avoid a type flip. Rejected because `mapFields()` prunes on `! empty( $value )` and an empty array prunes identically, so it buys nothing — while leaving `repeater` inconsistent with every other type, which returns `FALSE` here since #68.

## Environment

`parisek/timber-kit` v1.26.0, ACF Pro, WP 6.7, PHP 8.3, WPML. Reproduced on a `nav_menu_item` group (`nav_menu_item == location/main-menu`) with an unfilled `more_items` repeater.

## Aside, since it cost the investigation an hour

`formatFields()` resolves through `get_field_objects()`, which respects the group's location rule. Probing this under WP-CLI with a different theme active registers no menu locations, so every field reads as absent — indistinguishable from "the data is not there", with no warning. `get_field()` by name works in that same situation. Worth a line in the docblock; it is a convincing false negative.

parisek:
Closes #98.

## What changes

One condition in `Helpers::fieldFormatter()`:

```php
if ( ! isset( $field['value'] ) ) {
    if ( $is_preview && in_array( $field['type'], array( 'repeater', 'flexible_content' ), true ) ) {
        return $field;
    }
    return FALSE;
}
```

The pass-through is #68's block-preview contract and is kept — an unsaved block needs `sub_fields` / `layouts` to render a placeholder. It was ungated, so it also fired on ordinary front-end reads, where the definition array reads as a populated list: `{% if x and x|length > 0 %}` passes on the definition's own 28 keys, the template opens its wrapper, and the inner per-row guard suppresses every row. The result is empty chrome, not a visibly broken render — which is why it survived review. Measured on a `nav_menu_item` group: 67 of 69 items reported a 28-element `more_items`.

`$is_preview` was previously consulted only by the CF7 / WPForms branches, so this extends an existing convention.

## Tests

Test-first. The two tests #68 named are split by context rather than deleted:

| Test | Asserts |
| --- | --- |
| `test_repeater_null_value_returns_field_as_is_in_preview` | preview contract holds (now passes `true`) |
| `test_flexible_content_null_value_returns_field_as_is_in_preview` | same, flexible |
| `test_repeater_null_value_returns_false_outside_preview` | **new** — front end sees empty |
| `test_flexible_content_null_value_returns_false_outside_preview` | **new** — same, flexible |

Both new tests were watched failing against `main` (returned the definition array), then went green. `composer check` passes: 1036 unit tests, property suite, PHPStan level 8 clean, ADR index in sync.

## Behaviour change — declared

Public API unchanged (no signature, return type, or new argument). Exactly one input behaves differently: `repeater` / `flexible_content` with `value => null`, read with `$is_preview = false`. Before: field-definition array. After: `FALSE`, which `mapFields()` prunes, so the key is absent.

One regression path exists: a block render that bypasses `BlockRenderer` and omits `$is_preview` loses its editor placeholder — editor-only, front end unaffected, fixed by passing the flag. `BlockRenderer::buildContent()` (BlockRenderer.php:295) already propagates it correctly.

## Open point for review — the feature-flag doctrine

`AGENTS.md` § *Feature flags & breaking changes* says behavior-changing work ships behind a `StarterBase` flag, default off. This PR does not add one, and that is a judgement call worth an explicit yes or no:

- `Helpers` is static with no `StarterBase` instance in scope, so a flag would need a static property — a new mechanism, not the documented one.
- Defaulting it off would leave the definition-array leak on by default, i.e. keep the bug shipping.
- The surrounding block already states the invariant this restores: *"must read as empty, not leak the raw field-definition array into templates"*.

Treated as a bug fix rather than new behavior on those grounds. Happy to rework behind a flag if you read the doctrine as binding here.

## Rejected alternative

Returning `array()` instead of `FALSE` for the non-preview case. `mapFields()` prunes on `! empty( $value )`, so an empty array prunes identically — no benefit, while leaving `repeater` the one type not returning `FALSE` from this branch.

## Docblock

Two notes added while here: `$is_preview` on `formatFields()` now documents the placeholder role (it only mentioned shortcode suppression), and the location-rule false negative from the issue's aside is recorded — `get_field_objects()` honours the group's location rule, so probing a `nav_menu_item` group under WP-CLI with another theme active reads identically to "no data", with no warning. `get_field()` by name still resolves there.

https://claude.ai/code/session_01BsyBWmGD7itEupuAabxQC8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,15 +26,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   with nothing in it — which is why it survived review.
 
   Measured on a `nav_menu_item` group whose `more_items` repeater was filled on
-  10 of 425 items: 67 of 69 items reported a 28-element `more_items`.
+  10 of 425 items: 67 of 69 items reported a 28-element `more_items`. Options
+  pages resolve through the same gate, so an options-page repeater left unfilled
+  had the same symptom and is closed by the same change.
 
-  Outside preview the value is now `FALSE`, which `mapFields()` prunes, so the
-  key is simply absent. **Behaviour change, narrow:** the only affected input is
-  a `repeater` / `flexible_content` with `value => null` read with
-  `$is_preview = false`. `BlockRenderer::buildContent()` already propagates the
-  flag; a block render path that bypasses `BlockRenderer` and omits it would
-  lose its editor placeholder (editor-only, front end unaffected) and is fixed
-  by passing the argument the signature already carries.
+  Outside preview the value is now `FALSE`. **Where that lands differs by
+  depth:** `formatFields()` prunes on `! empty( $value )`, so a *top-level*
+  field drops out of the context entirely — the key is absent. A *nested*
+  repeater / flexible_content inside a populated parent row is assigned in
+  place by the recursion and is not pruned, so the row carries `links => false`
+  rather than omitting the key. Both read as empty in Twig; PHP consumers doing
+  `array_key_exists()` on a nested row see `false` where they previously saw the
+  definition array.
+
+  **Behaviour change, narrow:** the only affected input is a `repeater` /
+  `flexible_content` with `value => null` read with `$is_preview = false`.
+  `BlockRenderer::buildContent()` already propagates the flag; a block render
+  path that bypasses `BlockRenderer` — a legacy per-theme
+  `timber_block_render_callback()` predating the `BlockRenderer` migration — and
+  omits it would lose its editor placeholder (editor-only, front end
+  unaffected), and is fixed by forwarding the `$is_preview` WordPress already
+  hands that callback.
 
 - **Release guard runs every test suite.** `release-stamp.yml` ran
   `composer test`, which is the Unit suite only (`--testsuite=Unit`), so a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 
+- **`fieldFormatter()` gates the repeater/flexible null-value pass-through on
+  `$is_preview`** ([#98](https://github.com/parisek/timber-kit/issues/98)). An
+  unfilled `repeater` / `flexible_content` returned its raw ACF field-definition
+  array regardless of context. That is the block-preview contract — an unsaved
+  block needs `sub_fields` to render a placeholder — but it was ungated, so it
+  also fired on ordinary front-end reads, where the definition array reads as a
+  populated list: `{% if x and x|length > 0 %}` passes on the definition's own
+  28 keys, a template opens its wrapper, and the inner per-row guard then
+  suppresses every row. The visible result is empty chrome — a bordered `<ul>`
+  with nothing in it — which is why it survived review.
+
+  Measured on a `nav_menu_item` group whose `more_items` repeater was filled on
+  10 of 425 items: 67 of 69 items reported a 28-element `more_items`.
+
+  Outside preview the value is now `FALSE`, which `mapFields()` prunes, so the
+  key is simply absent. **Behaviour change, narrow:** the only affected input is
+  a `repeater` / `flexible_content` with `value => null` read with
+  `$is_preview = false`. `BlockRenderer::buildContent()` already propagates the
+  flag; a block render path that bypasses `BlockRenderer` and omits it would
+  lose its editor placeholder (editor-only, front end unaffected) and is fixed
+  by passing the argument the signature already carries.
+
 - **Release guard runs every test suite.** `release-stamp.yml` ran
   `composer test`, which is the Unit suite only (`--testsuite=Unit`), so a
   version could be stamped without the property suite's generative assertions.
--- a/src/Helpers.php
+++ b/src/Helpers.php
@@ -762,12 +762,21 @@ class Helpers {
 	 * contexts).  Each field value is passed through {@see fieldFormatter()}.
 	 * Fields with an empty formatted value are omitted from the result.
 	 *
+	 * Resolution goes through `get_field_objects()`, which honours the field
+	 * group's location rule. A group scoped to something the current request
+	 * does not establish — a `nav_menu_item` rule under WP-CLI with another
+	 * theme active, so no menu location is registered — surfaces no fields at
+	 * all, and reads identically to "the data is not there". `get_field()` by
+	 * name still resolves in that situation; prefer it when probing.
+	 *
 	 * @param object|int|string|null $post       Post object, term object, numeric post ID,
 	 *                                            options-page string key, or null to use the
 	 *                                            current queried object.
 	 * @param bool                   $is_preview True when rendering inside a Gutenberg block
-	 *                                            preview (suppresses shortcode execution for
-	 *                                            certain form plugins).
+	 *                                            preview. Suppresses shortcode execution for
+	 *                                            certain form plugins, and keeps the raw
+	 *                                            definition of an unfilled repeater /
+	 *                                            flexible_content so a placeholder can render.
 	 * @return array<string, mixed> Associative array keyed by ACF field name with formatted values.
 	 */
 	public static function formatFields( $post = null, $is_preview = false ) {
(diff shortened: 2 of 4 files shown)


parisek:
## Agent review — Codex (GPT-5.1-Codex, 2026-08-01) and Claude Fable 5 (2026-08-01)

Two reviewers, run independently with no sight of each other's output, both briefed to attack downstream upgrade regression risk specifically. Summarised, not pasted.

### Where they converged

Agreement between reviewers that cannot see each other is the strongest signal here, and they converged on three points:

1. **`$is_preview` propagation inside the package is faithful.** Both traced the full call graph independently and reached the same conclusion: `BlockRenderer::render()` → `buildContent()` → `formatFields($post_id, $is_preview)`, all four recursive calls (`Helpers.php:1263`, `:1272`, `:1302`, `:1308`) thread it unchanged, `formatMenu()` non-preview by design, and no block-render path in `src/` bypasses `buildContent()`. No finding.
2. **Coverage gap at the `formatFields()` level** — the fix was asserted only at `fieldFormatter()` in isolation, never at the level the bug was actually found. **Fixed** (see below).
3. **The one real downstream vector is a legacy `timber_block_render_callback()`** outside `BlockRenderer` that doesn't forward `$is_preview`. Both proposed the same mitigation: grep the fleet. **Done** (see below).

### Findings acted on

| # | Reviewer | Finding | Outcome |
| --- | --- | --- | --- |
| 1 | Codex | CHANGELOG's "the key is simply absent" holds only at the top level; the recursion assigns in place without pruning, so a nested unfilled repeater lands in its parent row as `false` | **Fixed** — verified in code first (`Helpers.php:837-839` prunes, `:1263` does not), CHANGELOG now states both depths |
| 2 | Both | Missing `formatFields()`-level tests, incl. recursion and preview propagation | **Fixed** — four tests added; the nested pair was mutation-checked (dropping `$is_preview` at `:1263` turns it red) |
| 3 | Fable | CHANGELOG credits the pruning to `mapFields()`, which does not exist in this package | **Fixed** — confirmed `grep` finds no such symbol in `src/`. The name came from #98 and I repeated it without checking; the mechanism is the inline `! empty( $value )` guard in `formatFields()` |
| 4 | Fable | Options pages resolve through the same gate, so the same latent bug is closed there too | **Adopted** — added to the CHANGELOG; it widens the fix's value rather than changing it |

### Verified and rejected

- **Fable independently re-derived the `array()` vs `FALSE` rejection** rather than trusting the PR text: `Helpers.php:837`, both satisfy `empty()`, pruning identical. Claim holds.
- **Fable traced the Twig idioms** (`and`, `is defined`, `|default([])`, `for`) and found every one an improvement, none a regression. The `is defined` case is the sharpest: previously always true for an unfilled repeater, so a bare `{% for %}` iterated the field's own metadata as rows.

### Fleet scan — the convergent mitigation

Both reviewers asked for it, so it was run across the 25 consuming projects (all on `^1.x`, so all pick this up on upgrade):

- **Twig `strict_variables`: not enabled anywhere.** This was the only candidate for hard breakage — a missing key would throw rather than read `null`. It does not exist in this fleet.
- **No template anywhere reads `.sub_fields` / `.layouts`** — nothing consumes the definition array.
- **PHP-side subscripting of `formatFields()` results is uniformly `?? ''` / `?? array()` guarded.**
- **The legacy-callback vector does not materialise.** Seven themes do carry a pre-`BlockRenderer` `classes/Block.php` with its own private `fieldFormatter()` — but that is the theme's own method, and the only `Helpers::` call in those files is `Helpers::formatImage()`. They never reach the changed branch.
- The idiom the fix corrects (`|length > 0` on a repeater-shaped key) appears **109 times** across the fleet — all currently latent instances of this bug, all moving from wrongly-true to correctly-false.

### Reviewer accuracy

Both reports held up against the code; every finding above was checked before acting on it. Codex's nested-`false` finding is the one I would have missed — it is correct and reveals that the top-level and nested cases genuinely differ. Fable's `mapFields()` catch is small but the useful kind: it caught a wrong function name propagating from the issue into the changelog unchecked.

🤖 review orchestrated with [Claude Code](https://claude.com/claude-code)

PR #101, opened 2026-08-02
feat(starter-base): allow $options_pages entries to set ACF post_id

### LINKED ISSUE #100: StarterBase: allow $options_pages entries to set ACF post_id (storage namespace)
## Summary

`StarterBase::$options_pages` lets a project configure `menu_slug`, `page_title`, `parent_slug`, `capability`, `icon_url` and `admin_bar`. It does not expose ACF's **`post_id`**, so `register_options_page()` builds:

```php
$args = [
    'page_title'      => $page['page_title'],
    'menu_title'      => $page['page_title'],
    'menu_slug'       => $page['menu_slug'],
    'capability'      => $page['capability'] ?? 'edit_posts',
    'redirect'        => false,
    'show_in_graphql' => false,
];
```

ACF therefore falls back to its default, `post_id = 'options'`, and every options page on the install shares one storage namespace: values land in `wp_options` as `options_<field_name>`, keyed by **field name, not by page**.

That is fine while a site has exactly one set of options pages — which is the assumption the current shape encodes, and it holds for most projects. It stops holding the moment a second options page exists, and nothing in the configuration says so, because the default never appears there.

## What it looks like when it breaks

Found on `sloneek` (WordPress, timber-kit v1.26.1) during a legacy migration.

The install carries a second options page created years ago through ACF's admin UI — `global-switchers`, a `acf-ui-options-page` post in the database, so it survives theme switches and does not appear in any `grep` of the repo. It also defaults to `post_id = 'options'`.

Both sides happen to own a field named `footer`. The result:

```
acf_get_field_groups(['options_page' => 'settings'])   → 5 groups, all active
Helpers::formatFields('option')                        → show_notification_bar_switcher,
                                                          sale_timer_global,
                                                          hide_notification_bar,
                                                          footer
```

The five groups the theme registers are correct and active, and `formatFields('option')` returns the *other* page's fields. The theme's own `links`, `announcement`, `social` and `exit_popup` groups never surfaced at all — and `footer` in that list is the legacy field, not the theme's group.

Two things make this expensive to find rather than merely wrong:

1. **A name collision reads as success.** The theme's footer template renders app-store buttons from `content.apps`; the legacy rows are stored as `options_footer_apps_*`, which is exactly what the theme's own `footer.apps` selector resolves to. So wiring the footer "worked", verifiably, against data that belonged to a different page. I recorded that render as evidence my change was correct. It wasn't.

2. **`get_field()` and `get_fields()` disagree silently.** `get_field('links_login', 'option')` returns the stored value; `get_field('links', 'option')` returns `null`; `formatFields()`, which assembles groups, sees nothing. No error, no warning, no log line — an empty options group is indistinguishable from one nobody filled in.

## Proposal

Accept an optional `post_id` on an `$options_pages` entry and pass it through:

```php
if ( isset( $page['post_id'] ) ) {
    $args['post_id'] = $page['post_id'];
}
```

A project that wants isolation then writes:

```php
protected array $options_pages = [
    [ 'menu_slug' => 'settings', 'page_title' => 'Theme Settings',
      'post_id' => 'sloneek_settings', 'admin_bar' => true ],
];
```

and its values are stored as `sloneek_settings_<field_name>`, which cannot collide with anything else on the install.

## Is it a breaking change?

**No, on its own.** Omitting the key leaves ACF's default in place, so every existing consumer behaves exactly as today. Purely additive to both the property shape and the generated `$args`.

**Yes, for whoever adopts it.** Setting `post_id` changes where values are read from and written to, so an existing site would need its stored options migrated to the new prefix — and reads go through `formatFields($post_id)`, which means the theme's own call site changes too. That is a per-project decision with a visible cost, which is an argument for making it opt-in exactly as proposed, not for changing the default.

One thing I could not settle from the outside: whether `acf_add_options_sub_page()` inherits a parent's `post_id` when the child omits it, or falls back to `'options'` independently. If the latter, a parent with a custom `post_id` and unmarked children would split storage across two namespaces — worth deciding deliberately rather than inheriting ACF's behaviour by accident.

## What this does not ask for

The colliding page on our install is a legacy artefact and removing it is our cleanup, not this package's problem. The request here is only that a project be *able* to isolate its own namespace, so the next install that grows a second options page fails loudly — or better, cannot collide at all.

## Environment

`parisek/timber-kit` v1.26.1, ACF Pro, WP 6.7, PHP 8.3, WPML.

parisek:
Closes #100.

## From project

`sloneek` (WordPress, timber-kit v1.26.1), during a legacy Sage → Timber migration. Options values migrated correctly into the database and nothing reached the front end.

## Why

ACF defaults every options page to `post_id = 'options'`, so values land in `wp_options` as `options_<field_name>` — **keyed by field name, not by page**. `$options_pages` exposed no way to change that, so a theme shared one storage namespace with every other options page on the install.

That assumption holds until a second options page exists, which can happen with no repo change at all: the sloneek install carries one created years ago through ACF's admin UI — an `acf-ui-options-page` post in the database, invisible to any grep.

Both sides owned a field named `footer`:

```
acf_get_field_groups(['options_page' => 'settings'])   → 5 groups, all active
Helpers::formatFields('option')                        → show_notification_bar_switcher,
                                                          sale_timer_global,
                                                          hide_notification_bar,
                                                          footer   ← the other page's
```

The theme's own `links`, `announcement`, `social` and `exit_popup` groups never surfaced. Two things made this expensive rather than merely wrong:

1. **A name collision reads as success.** The theme's footer rendered app-store buttons correctly — from `options_footer_apps_*` rows belonging to the *other* page, which happen to match the theme's own selector. That render was recorded as evidence the wiring was right. It wasn't.
2. **`get_field()` and `get_fields()` disagree silently.** `get_field('links_login', 'option')` returns a value, `get_field('links', 'option')` returns `null`, `formatFields()` sees nothing. No error, no log line — an empty options group is indistinguishable from one nobody filled in.

## What

An optional per-entry `post_id`, passed through to `acf_add_options_page()` / `acf_add_options_sub_page()`:

```php
protected array $options_pages = [
    [ 'menu_slug' => 'settings', 'page_title' => 'Theme Settings', 'post_id' => 'mytheme_settings' ],
    [ 'menu_slug' => 'footer', 'page_title' => 'Footer', 'parent_slug' => 'settings' ],
];
```

**A sub-page inherits its parent's `post_id`** unless it declares its own. This answers the open question in #100, and the answer is that ACF does *not* inherit: `acf_validate_options_page()` applies `'post_id' => 'options'` via `wp_parse_args` to every page independently, parent or not (verified against ACF Pro `pro/options-page.php`). Left alone, a namespaced parent with unmarked children would split one theme's settings across two namespaces and `formatFields()` would return only half of them — the same silent-half-empty failure this PR exists to end. Within a single `$options_pages` list a sub-page is visibly part of the same store, so it is treated as one; declaring `post_id` on the child opts out.

The **read side needed no change**: `Helpers::getFieldObjectsForOptions()` already filters `acf_get_options_pages()` by namespace and `decodeOptionsNamespace()` already collapses the `option`/`options` alias. The gap was only at registration.

## Breaking change?

**No, on its own.** Omitting the key leaves ACF's default in place, so every existing consumer behaves exactly as today. Additive to both the property shape and the generated `$args`.

**Yes, for whoever adopts it.** Setting `post_id` changes where values are read and written, so a live site needs its stored options copied to the new prefix, and the theme's own `formatFields()` call site changes too. That is an argument for opt-in — which is what this is — not for changing the default.

## Deliberately not done

- **No default change.** Namespacing by default would silently orphan every existing site's options.
- **No collision detection.** Warning about a shared namespace would require guessing which of two pages is the intruder; namespacing removes the collision instead of reporting it.
- **The colliding page itself** is the downstream project's cleanup, not this package's problem.

## Tests

`tests/Unit/StarterBase/AcfOptionsPagePostIdTest.php`, 7 cases asserting the `$args` handed to ACF — absent when undeclared, passed through, inherited by a sub-page, inherited regardless of declaration order, child override wins, no inheritance from an un-namespaced parent, no inheritance from a `parent_slug` pointing outside the list.

`composer check` green: 1041 unit + 8 property tests, PHPStan clean, ADR index OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,16 +22,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   to match the theme's own selector. A name collision reads as success, which is
   what made it expensive to find rather than merely wrong.
 
-  A sub-page **inherits its parent's `post_id`** unless it declares its own. ACF
-  does not do this — `acf_validate_options_page()` defaults every page to
-  `'options'` independently, parent or not — and without the inheritance a
-  namespaced parent with unmarked children would split one theme's settings
-  across two namespaces, with `formatFields()` returning only half of them. The
-  read side needed no change: `Helpers::getFieldObjectsForOptions()` already
-  matches on namespace.
+  A sub-page **inherits its parent's `post_id`** unless it declares its own, and
+  the inheritance is transitive across nesting levels. ACF does not do this —
+  `acf_options_page::validate_page()` applies `'post_id' => 'options'` through
+  `wp_parse_args` to every page independently, parent or not — and without the
+  inheritance a namespaced parent with unmarked children would split one theme's
+  settings across two namespaces, with `formatFields()` returning only half of
+  them. The read side needed no change: `Helpers::getFieldObjectsForOptions()`
+  already matches on namespace.
 
   Adopting it on a live site is a data migration — stored values stay behind
   under the old prefix.
+
+### Fixed
+
+- **Breeze cache purge missed options pages with a custom `post_id`.**
+  `clear_cache_on_options_save()` compared the incoming id against the literal
+  `'options'`, but ACF's admin controller saves through
+  `acf_save_post( $page['post_id'] )` — so a namespaced page (and, under WPML,
+  any language-suffixed id from `acf_get_valid_post_id()`) fired the hook with
+  an id the check rejected, and stale pages kept being served after a save. It
+  now matches any id that `acf_decode_post_id()` classifies as an options
+  namespace. The purge is site-wide either way, so widening the match can only
+  flush on another page's save — the safe direction for a cache.
 
 - **README badges** — Packagist version, PHP version, Timber, Tests, License.
   Matches `parisek/definition-kit` and `parisek/acf-json-schema`, which already
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- **README badges** — Packagist version, PHP version, Timber, Tests, License.
+  Matches `parisek/definition-kit` and `parisek/acf-json-schema`, which already
+  carried the same row; `parisek/styleguide` gains it in parallel.
+
 ### Changed
 
 - **Release guard runs every test suite.** `release-stamp.yml` ran
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- **`post_id` on an `$options_pages` entry** — sets the ACF storage namespace, so
+  a theme's options no longer share the default `options_<field_name>` prefix in
+  `wp_options` with every other options page on the install. Purely additive:
+  omit the key and ACF's own default applies exactly as before.
+  ([#100](https://github.com/parisek/timber-kit/issues/100))
+
+  Found on a site carrying a second options page created years earlier through
+  ACF's admin UI — a database row, invisible to any grep of the repo. Both sides
+  owned a field named `footer`; the theme's own `links`, `announcement`, `social`
+  and `exit_popup` groups never surfaced at all, while the footer rendered
+  *correctly* from the other page's data because `options_footer_apps_*` happened
+  to match the theme's own selector. A name collision reads as success, which is
+  what made it expensive to find rather than merely wrong.
+
+  A sub-page **inherits its parent's `post_id`** unless it declares its own. ACF
+  does not do this — `acf_validate_options_page()` defaults every page to
+  `'options'` independently, parent or not — and without the inheritance a
+  namespaced parent with unmarked children would split one theme's settings
+  across two namespaces, with `formatFields()` returning only half of them. The
+  read side needed no change: `Helpers::getFieldObjectsForOptions()` already
+  matches on namespace.
+
+  Adopting it on a live site is a data migration — stored values stay behind
+  under the old prefix.
+
 - **README badges** — Packagist version, PHP version, Timber, Tests, License.
   Matches `parisek/definition-kit` and `parisek/acf-json-schema`, which already
   carried the same row; `parisek/styleguide` gains it in parallel.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # timber-kit
 
+[![Packagist Version](https://img.shields.io/packagist/v/parisek/timber-kit.svg)](https://packagist.org/packages/parisek/timber-kit)
+[![PHP Version](https://img.shields.io/packagist/php-v/parisek/timber-kit.svg)](https://packagist.org/packages/parisek/timber-kit)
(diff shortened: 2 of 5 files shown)


parisek:
## Agent review — 2026-08-02

Two independent reviewers, neither able to see the other's output: **Codex** (`codex-rescue`) and **Fable**. Summarised, not pasted.

### Where they converged

**`acf_validate_options_page()` does not exist.** Both flagged it, cited three times in committed docs (docblock, README, CHANGELOG). Verified — `grep` of ACF Pro returns nothing. The real mechanism is the `acf_options_page::validate_page()` method applying `'post_id' => 'options'` via `wp_parse_args`, plus an `acf/validate_options_page` filter. The *behavioural* claim was correct; only the name was wrong. Fixed. Convergence on a factual detail is worth more than either report alone here — it is exactly the kind of thing one reviewer could plausibly hallucinate.

### Fixed

**Breeze purge missed custom namespaces** (Codex, medium). `clear_cache_on_options_save()` compared against the literal `'options'`, but ACF saves through `acf_save_post( $page['post_id'] )` — so adopting this PR's own feature silently disabled the cache purge, leaving stale pages served after every options save. Under WPML the language-suffixed ids from `acf_get_valid_post_id()` fail the same check. Now matches anything `acf_decode_post_id()` classifies as an options namespace. This is the most valuable finding of the two reports: the feature actively broke a neighbouring one, and no test covered the interaction.

**Inheritance stopped at one level** (Fable, medium). The parent map was seeded from top-level entries only, so a sub-page under a sub-page fell back to `'options'` while its siblings sat under the theme's namespace — this PR's exact failure class, one level deeper. Verified that ACF's `add_sub_page()` places no restriction on `parent_slug` pointing at another sub-page. Resolution now walks the hierarchy in repeated passes; a `parent_slug` cycle terminates rather than recursing.

**Docs.** `post_id` described as top-level-only (it is per-entry), and a pre-existing docblock line claiming `parent_slug` must reference a top-level page in the same list — contradicted by this PR's own foreign-parent test. Both corrected.

### Rejected

**Duplicate `menu_slug`** (Codex, medium). Two top-level entries sharing a slug and differing in `post_id`: ACF keeps the first, our map keeps the last, so a child could inherit the wrong namespace. Real, but it is a misconfiguration ACF already resolves silently and inconsistently on its own; validating it belongs with registration validation generally, not with this feature. Fable independently checked the same case and reached the same conclusion — not worth blocking on.

**Missing `menu_slug` / `page_title` validation** (Codex, low). Pre-existing: `$args['menu_slug'] = $page['menu_slug']` already warns on a malformed entry. The new map access adds no failure mode that did not exist.

**Read-side changes.** Both reviewers independently confirmed `Helpers::getFieldObjectsForOptions()` / `decodeOptionsNamespace()` need no change — they already resolve whatever namespace each page actually registered with.

Codex reported the docblock's `parent_slug` line as *contradicted by the tests* — correct — but framed the foreign-parent test as evidence the package "supports" external parents. It does not support them; it declines to inherit into them, which is a different claim. The docblock now says what actually holds rather than either version.

### Tests

7 → 13 cases in `AcfOptionsPagePostIdTest.php` (transitive inheritance, deepest-first order, mid-level override, cycle termination), plus 2 in `ClearCacheOnOptionsSaveTest.php`. `composer check` green: 1047 unit + 8 property, PHPStan clean.

Issue #103, opened 2026-08-04
formatFields() resolves taxonomy terms against post ids


parisek:
`formatFields()` assigns `$post->term_id` as a bare integer and passes it to `get_field_objects()`, which ACF interprets as a post id. A taxonomy term therefore resolves against the post carrying the same number — silently, returning the wrong fields or none at all. No warning, no exception; the caller simply gets a field map that looks plausible and belongs to something else.

ACF addresses a taxonomy term as the string `"term_<id>"`. The bare-integer form the id resolution produces can never reach one.

Found while adding menu-level metadata to `formatMenu()` (#66). That work needed ACF fields attached to the `nav_menu` term, and could not use `formatFields()` for exactly this reason — it ships `getFieldObjectsForNavMenu()`, which hand-builds the ACF screen and passes `"term_<id>"`, working the problem around narrowly for `nav_menu` terms only.

The general case is still affected: `formatFields($category)` and any other taxonomy term.

Worth noting the kit already carries the neighbouring workaround one level down — `isNavMenuItemPostId()` + `getFieldObjectsForNavMenuItem()` exist because ACF's screen detection is incomplete for `nav_menu_item` posts. So there are now two hand-built escapes around the same gap, which is an argument for fixing the id resolution itself rather than adding a third.

Fixing it changes `formatFields()` behaviour for every caller, so it wants its own release note — and probably its own decision about whether the fix is a `term_` prefix at the resolution site or a dedicated taxonomy branch mirroring the existing menu-item one.

Issue #66, opened 2026-07-04
timber-kit 2.0 — planning meta-issue


parisek:
Meta-issue tracking everything queued for the next major. Source: June 2026 review of the kit against 20 production projects + `starter_theme` (all Porta Design WP sites consuming `parisek/timber-kit`). Each item should land as its own PR/discussion; check off here.

Context from the audit, for calibration: 21/22 installs lag behind the latest release (11 frozen on v1.7.7), zero projects patch or fork the kit — all friction is expressed as `Base.php` overrides. 2.0 is the one chance to break things deliberately; everything below is either impossible or awkward under the 1.x compatibility contract.

## Deprecated removals (blocked until 2.0, already documented)

- [ ] Remove `Helpers::resizeImage()` (deprecated in favor of `Resizer` / `ImageHelper::resize()`; ~100 lines + its test file)
- [ ] Remove `StarterBase::$max_upload_width` / `$max_upload_height` (superseded by `$big_image_size_threshold`; only one audited project — aleszejdl — still sets them)
- [ ] Remove `StarterBase::resize_uploaded_image()` (no longer hooked; docblock-only deprecation)

## Filter naming unification

- [ ] Standardize on the underscore convention `timber_kit_*` (14:1 majority) and migrate the slash-style `timber_kit/block_renderer/*` / `WpmlBlockOverride` filters. Ship a thin alias layer (old name proxies to new, `_deprecated_hook()` notice) for one transition release, then drop. The draft breadcrumb ADR already flagged this inconsistency as known-but-unresolved.

## Flag default flips (breaking by doctrine, so 2.0-only)

- [ ] `$wpml_block_override`: default `false` → `true`? Enabled by 7/20 projects + `starter_theme`, always via the same copy-pasted "Porta Design best-practice" block. Decide, and if flipped, document the opt-out.
- [ ] `$acf_datastore`: default `false` → `true`? Same best-practice block, 5/20 + `starter_theme`. Blocked on WPML certification noted in its docblock.
- [ ] `$disable_search`: keep default `true` but document loudly — 5/20 projects explicitly re-enable search, and 7 more set `true` redundantly (a sign the default is invisible). Alternative: no flip, but add a commented flag inventory to `starter_theme` so every scaffold sees the defaults.

[...]

PR #104, opened 2026-08-04
feat(helpers): expose menu-level metadata from formatMenu()

parisek:
## What

`Helpers::formatMenu()` returned a flat list of items and discarded everything about the menu itself. It now returns a `MenuData` object exposing the menu's own `id`, `title`, `name`, `slug` and `description` alongside `items` — plus any ACF fields attached to the `nav_menu` term.

## Why

A footer column in mairateam had its heading hardcoded in Twig (`_x('Maira', …)`). The client needs "Kanály" in Czech and "Channels" in English, and because the string lives in the template it is reachable neither by an editor nor by WPML String Translation. Taking the heading from the menu name solves both languages at once, since WPML already assigns a different menu per language to the same location.

The wider goal is the ACF surface: a `nav_menu` is a `WP_Term`, so a project can now attach a field group to the Menu location and give a menu an icon, a colour or a visibility flag **without a kit release**.

## Backward compatibility

An audit across the 26 themes consuming this package:

| Pattern | Count | Affected |
| --- | ---: | --- |
| `'key' => Helpers::formatMenu('slug')` into the Timber context | 140 | no |
| Twig `{% if …menu… %}` truthiness guards | 461 | no |
| Twig `\|merge`, `\|slice`, `\|length` | 6 | no |
| PHP `is_array()` / `array_map()` over a menu | 0 | — |
| `json_encode()` over a menu, direct or via `data-*` | 0 | — |

**No call site needs to change.**

`MenuData` implements `IteratorAggregate`, `ArrayAccess`, `Countable` and `JsonSerializable`, so iteration, `count()`, index access and JSON encoding behave exactly as on the array it replaced.

### The one deliberate wart

`formatMenu()` returns **two types**: a plain `[]` for an empty or missing menu, `MenuData` otherwise.

This is not an oversight. A PHP object is always truthy — `Countable` does not affect the boolean cast — so returning an empty `MenuData` would silently flip all 461 truthiness guards and render empty `<nav>` columns across every consuming site. Returning `[]` keeps the empty case falsy, iterable and zero-count, exactly as today.

Normalising to a single return type is queued for 2.0, recorded on #66.

Sub-levels (`item.below`) also stay plain arrays: they are not menus, carry no term metadata, and wrapping them would change `below` for every consumer.

## The ACF resolver

`Helpers::formatFields( $menu )` could not be used. Its id resolution hands a **bare integer** to `get_field_objects()`, which ACF reads as a *post* id — a term would silently resolve against an unrelated post. ACF addresses a taxonomy term as `"term_<id>"`.

This ships a narrow `getFieldObjectsForNavMenu()` mirroring the existing `getFieldObjectsForNavMenuItem()` workaround. The general bug affecting every taxonomy term is **not** fixed here — it changes behaviour for every caller and is filed separately as #103.

## Tests

`formatMenu()` was the only `Helpers::format*` function with no coverage. Four layers, each answering a different question:

1. **Baseline** — locks the pre-change item shape, `below` recursion and the WPML description branch, so the change has something to regress against.
2. **Observational equivalence** — asserts `MenuData` is indistinguishable from the array across every axis a consumer can observe. Adds no functionality; it is the layer that substantiates the zero-migration claim.
3. **Twig integration** — the conclusion that `|merge` and `|slice` accept a `Traversable` came from reading Twig's `CoreExtension`, not running it. These render real templates lifted verbatim from the audited call sites (keypers `|merge`, anyever `|slice`, eprukaz/oekoplan `|length`, mairateam guard + iteration) against Twig 3.27.1. **Both pass.**
4. **Property-based** — the equivalence invariants across generated inputs, with the empty/non-empty boundary deterministically exercised.

1092 unit tests, 15 property tests, PHPStan clean.

## Verified against the real consumer

Against a copy of the production database, the footer heading rendered `Footer Kanály` from the menu name, and changed to `Kanály` when the term was renamed.

The English column is **not** verified: it needs a WPML-paired EN menu that does not exist yet. That is the admin half of the originating ticket, not a kit concern. Worth noting the empty-menu design behaved correctly there — no EN menu produced `[]`, so the guard hid the column instead of rendering an empty `<nav>`.

## Notes for review

- Test `setUp()` methods gained defensive `is_plugin_active` / `acf_get_field_groups` / `get_field_objects` stubs. Brain\Monkey patches functions process-wide, so once any earlier test file mocks one, `function_exists()` returns true for every later file. These are order-independence hardening, not scope creep.
- An ACF field on the `nav_menu` term named `id`, `title`, `name`, `slug`, `description` or `items` is silently dropped by the merge. Documented in the README.

Ships as a **minor** (1.28.0) — purely additive.

https://claude.ai/code/session_01Xhu5oYcJS67XGTJ1tebmM5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- **Menu-level metadata from `formatMenu()`** — the returned value now carries the
+  menu's own `title`, `name`, `slug`, `description` and `id`, plus any ACF fields
+  attached to the `nav_menu` term. A footer column can take its heading from the
+  menu name instead of a hardcoded template string, which also makes it
+  translatable through WPML's normal per-language menu assignment.
+
+  Backward compatible by construction: the return value is a `MenuData` object
+  that iterates, counts, indexes and JSON-encodes exactly as the item list it
+  replaced, and empty or missing menus still return a plain `[]` so template
+  truthiness guards keep their meaning. An audit of 26 consuming themes found
+  140 call sites, 461 `{% if %}` guards and 6 filter uses — none require a change.
+
+  New ACF surface: attach a field group to the **Menu** location and its fields
+  appear as properties on the menu (`{{ menu.menu_icon }}`). This needed a
+  dedicated resolver — ACF addresses a taxonomy term as `term_<id>`, which the
+  generic `formatFields()` id resolution does not produce.
+
 ## [1.27.0] - 2026-08-02
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Static methods for formatting ACF data into clean arrays for Twig templates:
 - `formatFields()`, `fieldFormatter()` — ACF field processing
 - `formatLink()` — link/button formatting
 - `remapWpmlReference( $value, array $field, string $target_lang )` — remaps an ACF reference field's id(s) to a target WPML language via `wpml_object_id`, with the element type resolved per ACF field type (`image`/`file`/`gallery` → attachment, `post_object`/`relationship`/`page_link` → post, `taxonomy` → term; non-reference and non-numeric values pass through). Shared formatting-layer primitive that `WpmlBlockOverride` delegates to, reusable by any field formatter
-- `formatMenu()` — navigation menus
+- `formatMenu( $menu_or_name )` — navigation menus. Returns a `MenuData` object exposing the menu's own metadata (`title`, `name`, `slug`, `description`, `id`) alongside `items`, **plus any ACF fields attached to the `nav_menu` term** — so a project can give a menu an icon, a colour or a visibility flag without a kit release. The object iterates, counts, indexes and JSON-encodes exactly as the plain item list it replaced, so existing `{% for item in menu %}` call sites need no change. An empty or missing menu returns a plain `[]`, keeping `{% if menu %}` guards falsy. Note: `id`, `title`, `name`, `slug`, `description` and `items` are reserved metadata keys — an ACF field on the `nav_menu` term with one of those names is silently dropped in favour of the built-in value
 - `formatTerms()` — taxonomy terms
 - `formatLanguageSwitcher()` — WPML language switcher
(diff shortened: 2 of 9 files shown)


parisek:
Follow-up on the editorial side, filed against the MCP plugin now that this PR makes the menu name public-facing content:

- portadesign/portadesign-mcp#186 — `update-menu` owns items but not the menu itself: no rename, no create, no location assignment. Directly relevant here: with this PR merged, an editor can fill the column's links over MCP but cannot change the heading it renders.
- portadesign/portadesign-mcp#187 — no way to set a menu's language or pair it as a WPML translation, which is why the English column in the first consumer still cannot be completed.

Neither blocks this PR. Both are consequences of it worth tracking: the heading became editor-controlled content here, and the tooling an editor would reach for does not yet reach it.

parisek:
## ACF-on-`nav_menu` path verified against real ACF

Closing a gap worth naming: until now `getFieldObjectsForNavMenu()` was covered only by Brain\Monkey mocks. A mock answers what it was told to answer, so it could confirm the screen key `nav_menu` and the `"term_<id>"` value id only in the sense of confirming my own assumption. The mairateam production-data run did not close this either — that site has **zero** ACF field groups targeting `nav_menu`, so the ACF branch never executed there.

Verified directly against a live ACF Pro install (mairateam local, production DB copy, this branch patched into `vendor/`):

```php
acf_add_local_field_group([
    'key'    => 'group_probe_navmenu',
    'fields' => [[ 'key' => 'field_probe_menu_icon', 'name' => 'menu_icon', 'type' => 'text' ]],
    'location' => [[[ 'param' => 'nav_menu', 'operator' => '==', 'value' => 'all' ]]],
]);
update_field( 'field_probe_menu_icon', 'ico-probe-ok', 'term_49' );

$menu = Helpers::formatMenu( 'footer-secondary' );
```

Result:

```
typ návratu: Parisek\TimberKit\MenuData
title: 'Kanály'
menu_icon PŘES formatMenu: 'ico-probe-ok'
```

So both halves hold against real ACF: `acf_get_field_groups([ 'nav_menu' => $id ])` matches the Menu location rule, and `"term_<id>"` retrieves the value. The extensibility path the README now advertises is measured, not inferred.

The probe was removed and the vendored package restored afterwards; nothing was left behind.

Issue #106, opened 2026-08-04
acfml-sync-preferences: add --prune to remove orphaned repeater-index keys


parisek:
Sibling of #102. That one is about Copy-field **values** drifting apart; this one is about the **dictionary** that governs them growing without bound.

## Problem

`wp timber-kit acfml-sync-preferences` reconciles WPML's `custom_fields_translation` dictionary by walking existing postmeta and registering each key with its ACF field's `wpml_cf_preferences`. `PreferenceSyncPlan::patch()` returns additions and corrections only — there is no branch that removes anything. The dictionary is therefore monotonic: it only ever grows, on every deploy, for the life of the site.

That would be harmless if entries were per *field*. They are not. WPML looks custom fields up by **exact meta key**, and ACF stores repeater and flexible-content sub-fields flattened with the row index in the key. `WPML_ACF_Field_Settings::save_field_settings()` writes under `$field['name']`, which at save time is that flattened runtime name. So one `wpml_cf_preferences` declared once in `acf.json` materialises as one dictionary entry **per row index that has ever existed**, and nested repeaters multiply:

```
flat_detail_blocks_0_items_0_image
flat_detail_blocks_0_items_1_image
flat_detail_blocks_1_items_0_image
…
```

Delete the row and the entry stays. Add a fourth item to one repeater on one page and the index-3 keys are registered forever.

### Measured on a live site

Composition of the dictionary's Copy list on `fellows`:

| Key shape | Count | Share |
| --- | ---: | ---: |
| repeater-index (`…_0_…`) | 353 | 86 % |
| `_<name>` companions | 36 | 9 % |
| plain field names | 23 | 5 % |

Only 23 entries out of 412 correspond to something a human declared. The staging copy of the database has 412 Copy keys; production has **807** — the same schema, twice the content history.

The stored option is **99.4 KB and autoloaded**. An earlier version of this issue claimed that cost is paid on every request; that was asserted, not measured, and it is wrong. Loading and unserialising it, and calling `get_custom_fields_translation_settings()`, both measure at ~0.0000s. On ordinary site traffic a large dictionary is very nearly free — only `copy_custom_fields()` (linear, on translation-job save) and WPML's nav-menu sync walk the list at all.

### Why it stopped being theoretical — and why that turned out to be a different bug

An importer on that site synced Copy fields via `wpml_sync_all_custom_fields`, which walks the whole Copy list per post. Measured per flat, in loop conditions, on production:

| Flat | Cost of that one call |
[...]

PR #107, opened 2026-08-04
fix(wpml): sweep ACFML's acf/load_reference leak when syncing Copy fields

parisek:
## Why

Syncing Copy-marked custom fields post by post got slower with **every post already processed**. Measured on a live site (147 flats, 807 Copy keys), a single import that changed one field took 26 minutes.

The cause is not the database and not the posts. Query count stays flat at 30 per post while time grows 22x, and memory climbs. The same post costs **0.092s alone in a fresh process** and **4.64s at ordinal 40**.

`acfml` 2.2.4, `classes/class-wpml-acf-worker.php`:

```php
add_filter( 'acf/load_reference', $keepOnlyLastFieldReference );
$field = get_field_object( $metaKey, $objectFromId, false, false );
remove_filter( 'acf/$metaKey', $keepOnlyLastFieldReference );   // single-quoted
```

`$metaKey` is not interpolated, so the removal targets a literal hook name that never exists and the closure is never removed. ACFML runs that method **once per synced meta key**, so one `wpml_sync_all_custom_fields` over 807 Copy keys leaks 807 callbacks — confirmed by snapshotting `$wp_filter` around a single sync:

```
post 1220 — hooks gained by one sync_all:
  acf/load_reference   +807
      last: Closure wp-content/plugins/acfml/classes/class-wpml-acf-worker.php:97
```

## What

- **`Acfml\LoadReferenceGuard`** — removes callbacks added to `acf/load_reference` since its construction, restoring the state ACFML itself intends (the closure is scoped to one `get_field_object()` call and has no purpose after it).
- **`Wpml\CopyFieldSync`** — the sanctioned way to push Copy fields in a loop: sync only the keys the caller wrote, then sweep. Handles `_<name>` companions.
- README section under *ACFML preference sync*, CHANGELOG entry, `WP_Hook` test-bootstrap stub alongside the existing `wpdb`/`WP_Error` ones.

24 posts, 807 Copy keys, one process:

| | first post | last post | total |
|---|---|---|---|
| as shipped | 0.060s | 1.327s (22x) | 16.33s |
| swept after each post | 0.047s | 0.040s (0.8x) | **0.95s** |

The flat curve matters more than the 17x: cost stops depending on the iteration ordinal.

## Deliberately not done

- **No `wp_cache_flush()` between posts.** First hypothesis, measurably wrong — the leak lives in `$wp_filter`, not the object cache. Flushing left growth at 2.3x and pushed the total from 16.33s to 52.05s by re-querying everything.
- **No dictionary pruning here.** That treats the multiplier, not the mechanism; with the leak swept, syncing all 807 keys costs 0.040s/post. Pruning stays worth doing as hygiene (#106) but is no longer urgent.
- **No version check on acfml.** Once upstream fixes the typo, `sweep()` finds nothing and returns 0. Behaviour keyed to a plugin version would be one more thing to get wrong.
- **Named callbacks are never removed** — only closures. A plugin registering a named function mid-loop is a deliberate act; every leaked callback is anonymous, so the narrower rule loses nothing.

Reported upstream to WPML separately.

## Verification

`composer check` green — 1112 unit + 15 property tests, PHPStan level 8 clean, ADR index OK. 14 new tests cover baseline scoping, priority coverage, the named-callback exclusion, `around()` sweeping on throw, and the post-upstream-fix no-op case.

https://claude.ai/code/session_01SKiM9NkCH6fCX4rC1Wuq32

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- **`Acfml\LoadReferenceGuard` + `Wpml\CopyFieldSync`** — sync Copy-marked
+  custom fields to translations in a loop without the two costs that make the
+  obvious implementation unusable.
+
+  ACFML 2.2.4 leaks a filter callback per synced meta key:
+  `WPML_ACF_Worker::getFieldObjectWithFilteredReference()` adds a closure to
+  `acf/load_reference` and then removes `'acf/$metaKey'` — single-quoted, so a
+  literal hook name that never exists. Every later `get_field_object()` walks
+  the accumulated list, and the per-post cost rises with the iteration ordinal:
+  the same post measured 0.092s alone in a fresh process and 4.64s at ordinal
+  40. Sweeping restores what ACFML itself intends, so the guard becomes a no-op
+  once upstream corrects the typo.
+
+  `CopyFieldSync::push()` additionally syncs only the named keys instead of the
+  whole `custom_fields_translation` dictionary, which is site-wide, monotonic
+  and quadratic to walk. Over 24 posts with 807 Copy keys the combination went
+  from 16.33s to 0.95s and — more to the point — flattened the growth curve
+  from 22x to none.
+
 ## [1.28.0] - 2026-08-04
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -342,6 +342,59 @@ Applying newly-translatable keys triggers WPML's ProcessNewTranslatableFields
 background task — affected translations get flagged as needing update, which
 is the point: translators see the previously invisible backlog.
 
+### Syncing Copy fields in a loop
+
+**Never call `wpml_sync_all_custom_fields` per post in a loop.** It walks the
+whole dictionary each time, and that dictionary is site-wide and monotonic —
+ACFML registers one entry per *flattened* meta key, so a repeater contributes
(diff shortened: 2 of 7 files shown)


parisek:
Closing unmerged — the workaround is not worth carrying upstream.

**Fixed by WPML already.** ACFML 3.0-b.1 (shipped with the WPML 5.0 betas) has the correct line:

```php
add_filter( 'acf/load_reference', $keepOnlyLastFieldReference );
$field = get_field_object( $metaKey, $objectFromId, false, false );
remove_filter( 'acf/load_reference', $keepOnlyLastFieldReference );
```

Scope across our 47 ACFML installs:

| ACFML | State | Projects |
|---|---|---:|
| ≤ 2.1.5 | method does not exist yet | 17 |
| 2.2.2 – 2.2.4 | leak present | 30 |
| 3.0-b.1 (WPML 5.0 beta) | fixed | 1 |

And the leak only bites when the sync runs in a loop. The sole caller of `wpml_sync_all_custom_fields` outside WPML's own plugins across the whole workspace is `portadesign-mcp` (`Write/PostCreator.php`), which calls it **once per created post per request** — the leaked callbacks die with the request. The one place it mattered was a 147-post import on `fellows`, which is being fixed locally there.

So a permanent guard in timber-kit would be a cost with no beneficiary, and it would outlive the bug it guards against. The analysis is kept in #106 for anyone who meets this on a site still on 2.2.x; the branch `fix/acfml-load-reference-leak` stays if the code is ever wanted.

PR #108, opened 2026-08-05
feat(starter-base): wire per-language typography via a locale resolver

parisek:
## Summary

`parisek/twig-typography` 1.3.0 gave `TypographyExtension` a second, optional
constructor argument: a zero-argument locale resolver invoked on every
`|typography` call. Without it, the new per-language typographic tables
(quote style, dash convention, single-character word spacing, ...) stay
inert and only the language-neutral house defaults apply.

- `timber_twig()` now passes `StarterBase::typography_locale_resolver()` as
  that second argument.
- `typography_locale_resolver()` is a new **overridable** method that
  delegates to the kit's existing `Helpers::getLanguage()` (WPML
  `wpml_post_language_details` / `wpml_current_language` filters, falling
  back to `get_locale()`) rather than duplicating that probe logic. A theme
  whose language detection doesn't go through WPML can override the method
  without forking `timber_twig()` — the same treatment `$typography_config`
  already gets for the settings-file path.
- Bumped `parisek/twig-typography` from `^1.0` to `^1.3` — passing a second
  constructor argument to 1.0-1.2 fatals with `ArgumentCountError`, so the
  floor has to rise with the new call shape.

## Design decision: `get_locale()`, not `determine_locale()`

The resolver's fallback (inside `Helpers::getLanguage()`, unchanged by this
PR) is `get_locale()`. I considered `determine_locale()` and decided against
it — the reasoning is written inline as a comment at the
`typography_locale_resolver()` call site:

Both functions substitute the *admin user's own* locale while `is_admin()`
is true, which is already the wrong signal for "language of the content
being rendered" (a Czech editor previewing an English post would get Czech
typography rules either way). `determine_locale()` goes further, though: it
also substitutes the user's locale for **any JSON request**
(`wp_is_json_request()`) — and the Gutenberg block-preview endpoint
(`wp-json/wp/v2/block-renderer/...`) is exactly that: a JSON request with
`is_admin()` false, hit while previewing a block whose post language may
differ from the logged-in editor's own UI language. `get_locale()` doesn't
special-case JSON requests, so on that request it still reflects the
site/post's own locale — and stays reachable by WPML's `locale` filter,
which `determine_locale()`'s early-return branches bypass entirely. Outside
admin/JSON requests the two functions agree, so this only diverges on
admin-side previews, which is exactly where getting it wrong would be most
visible.

## Other things checked

- **Second construction site?** Searched — `TypographyExtension` is
  constructed in exactly one place (`StarterBase::timber_twig()`). No other
  call site to update.
- **Existing test coverage of `timber_twig()`'s extension registration?**
  None existed before this PR (no test previously called `timber_twig()` and
  inspected registered extensions). Added
  `tests/Unit/StarterBase/TypographyLocaleResolverTest.php`, which unit-tests
  `typography_locale_resolver()` directly (WPML-filter path, `get_locale()`
  fallback path, and non-memoization across calls) rather than trying to
  reach into Twig's extension internals.

## Verification

- `composer check` (phpunit unit + property suites, phpstan, ADR index) —
  green.
- Manually rendered the same string through
  `TypographyExtension::applyTypography()` under two resolver-returned
  locales, proving the resolver is actually consulted:

  ```
  cs => On rekl „ahoj“ – jak se mas?
  en => On rekl "ahoj" — jak se mas?
  ```

  Czech gets low-quotes + spaced en-dash; English gets curly quotes + em-dash
  — same input string, different output, driven purely by the resolver's
  return value.

## Test plan

- [x] `composer check` green locally
- [x] New unit test covers the resolver's WPML + `get_locale()` fallback
      paths and non-memoization
- [ ] CI green on the PR


parisek:
Addressed all three review points:

1. **Region-subtag truncation fixed.** `Helpers::getLanguage()`'s `get_locale()` fallback no longer does `substr(..., 0, 2)` — it now returns the locale as `get_locale()` gives it (lowercased/trimmed), e.g. `de_CH` → `de_ch`, `fr_FR` → `fr_fr`. `LocaleResolver::candidates()` already parses both `_` and `-` forms and degrades to the base language on its own, so the truncation was pure information loss. The WPML branch above it is unaffected and untouched — confirmed WPML's `wpml_current_language` typically returns a bare language code (`cs`, `de`), so that's the path most WPML sites actually take; this fallback only fires when WPML is inactive or silent. Test updated (`GetLanguageTest`) plus a new regression case for the region-subtag survival, and manually verified end-to-end against the real package (couldn't run this inside PHPUnit — see note below):

   ```
   de_CH => Er sagte «Hallo»
   de_DE => Er sagte „Hallo"
   ```

2. **Docblock corrected** to quote WP core's actual `determine_locale()` condition (`is_admin()` OR a JSON request carrying `?_locale=user` — not "any JSON request") and to state precisely why `get_locale()` doesn't share that substitution.

3. **CHANGELOG corrected** — the floor bump is because 1.0–1.2 silently discards the new argument (PHP only throws `ArgumentCountError` on *internal* functions), not because it fatals. Silence, not noise, is the actual risk.

**Also done:**
- `typography_locale_resolver()` is now `protected` (override point for a `Base` subclass, not a Twig-callable like the neighboring `twig_*` methods — reasoning is in its docblock). Test now exercises it via `ReflectionMethod`, mirroring the existing `Breadcrumb` pattern.
- Added `test_timber_twig_registers_typography_extension_with_the_resolver`, which calls `timber_twig()` against a real `Twig\Environment` and reads `TypographyExtension`'s private `localeResolver` property via reflection to prove the resolver is actually wired through (not just defined).

**One thing I couldn't add to the suite:** an in-PHPUnit end-to-end render test (`de_CH` → guillemets) turned out to hang — `PHP_Typography`'s DOM walker passes nodes by reference internally, and Brain\Monkey's Patchwork interceptor turns that into a "must be passed by reference, value given" warning storm that never completes for real prose (confirmed: killed after 90s+, thousands of warnings). Same class of Patchwork/vendor-internals conflict the Property suite already works around by staying Brain\Monkey-free — not an option here since the resolver test needs `get_locale()`/`apply_filters` mocking. Documented the manual verification (above) as a comment in the test file instead of forcing a flaky/hanging in-suite test.

`composer check` green (1104 unit + 15 property tests, PHPStan clean, ADR index in sync). Commit `5b19ae4`.

PR #110, opened 2026-08-09
fix(breadcrumb): read the listing links from the namespace the options page registered

parisek:
## The bug

`Breadcrumb::get_global_links()` reads

```php
$links = get_field( 'links', 'option' );
```

unconditionally, while `StarterBase::register_options_page()` honours an explicit `'post_id'` in `$options_pages`:

```php
protected array $options_pages = [
    [ 'menu_slug' => 'settings', 'page_title' => 'Theme Settings', 'post_id' => 'mytheme_settings' ],
];
```

So the package **writes the listing links to one store and reads them from another**. `get_field()` returns `null`, `build_for_singular()` appends no listing item, and every singular of a `list_page_map` post type renders a trail one step short.

**Nothing reports it.** No error, no log line, no empty output — the trail still renders, just shorter.

## How it surfaced

A five-language WPML site whose seven listing links were all populated and none of which reached the breadcrumb. It read as "the option fields were never filled in", which was checked first and was not the cause.

What made it hard to see: a project-side `timber_kit_breadcrumb_items` filter injects the post's term, so the trail looked complete —

```
Úvod > AI > AI a automatizace v HR…          ← term in the slot, listing absent
```

Proven by probing the two stores directly:

```
get_field( 'links', 'option' )            => NULL
get_field( 'links', 'sloneek_settings' )  => article_list, blog_list, case_studies_list, …
```

## The fix

- `Breadcrumb` gains an `options_post_id` config key, **default `'option'`** — a project that never namespaced its options page is byte-identical.
- `StarterBase` passes the namespace it actually registered, derived from the first `$options_pages` entry declaring a `post_id`.
- `$breadcrumb_options_post_id` overrides that, for a theme whose `links` group lives on a page other than the first.

## Left alone on purpose

- **The group selector stays the literal `'links'`.** Making it configurable widens `list_page_map`'s contract and no consumer needs it.
- **`is_options_post_id()` / `acf_decode_post_id()` were not reused.** They answer "is this an options id", not "which namespace does this theme use", and routing through them would make the resolver depend on ACF being loaded — the breadcrumb has to degrade without it.
- **Two pre-existing failures on `main` are untouched**: PHPStan `TypographyExtension` constructor arity (`StarterBase.php:1363`) and PHPUnit `TypographyLocaleResolverTest`. Both reproduce on a clean checkout of `8a61223`; fixing them here would blur which failures this branch owns.

## Verification

| | before | after |
|---|---|---|
| `composer test` | 1112 tests, 1 error (pre-existing) | **1117 tests, 1 error** (same one) |
| `composer phpstan` | 1 error (pre-existing) | **1 error** (same one) |

Five new tests: the configured namespace is the one `get_field()` receives; the default is still `'option'`; and the three `StarterBase` resolution paths (derived / defaulted / explicitly overridden).

Downstream, on the site that surfaced it — all four singular types, all five languages:

```
cs  Úvod            > Blog > AI                 > AI a automatizace v HR…
en  Home            > Blog > AI                 > AI and Automation in HR…
sk  Úvod            > Blog > AI                 > AI a automatizácia v HR…
pl  Strona główna   > Blog > AI                 > AI i Automatyzacja w HR…
it  Home            > Blog > Consigli e trucchi > Employee Absence Software…
```

Three steps before, four now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,34 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **The breadcrumb's listing step is no longer dropped on a theme that
+  namespaces its ACF options page.** `Breadcrumb::get_global_links()` read
+  `get_field( 'links', 'option' )` unconditionally, while
+  `StarterBase::register_options_page()` honours an explicit `'post_id'` in
+  `$options_pages` (e.g. `'mytheme_settings'`) and writes there. The package
+  therefore wrote the listing links to one store and read them from another:
+  `get_field()` returned `null`, `build_for_singular()` appended no listing
+  item, and every singular of a `list_page_map` post type rendered a trail one
+  step short — with nothing logged and no error anywhere.
+
+  `Breadcrumb` gains an `options_post_id` config key (default `'option'`, so
+  un-namespaced projects are byte-identical), and `StarterBase` passes the
+  namespace it actually registered, derived from the first `$options_pages`
+  entry declaring a `post_id`. Override with `$breadcrumb_options_post_id` when
+  the `links` group lives on a different page than the first.
+
+  Found downstream on a five-language WPML site whose seven listing links were
+  all populated and none of which reached the breadcrumb. The failure is easy
+  to misread as "the fields are empty": a term filter had been supplying a
+  middle step, so the trail looked complete (`Home > AI > title`) while the
+  listing step it should have carried was silently absent.
+
+  **Not fixed here, deliberately:** the `links` group is still addressed by the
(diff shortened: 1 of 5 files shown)


parisek:
Correction to the "Left alone on purpose" section: **CI is green on all six checks**, including PHPStan and both PHP versions.

The two failures I described as pre-existing reproduce on my local checkout only — they come from local `vendor/` state (a `parisek/twig-typography` older than the one CI installs), not from `main`. The claim that they "reproduce on a clean checkout of 8a61223" is true of my machine and false of a clean install; I should have checked that before writing it.

Nothing in the diff changes as a result — the branch touches neither file — but the PR body's framing was wrong, and the verification table's "1 error before, 1 error after" describes my machine. On CI it is 0 before and 0 after.

PR #111, opened 2026-08-11
feat(outage): wp timber-kit outage-screen — install the drop-ins that serve it

parisek:
Design and consumer: portadesign/tailwind-base#569. Renderer: parisek/styleguide#124.

## What this adds

`wp timber-kit outage-screen install|status|remove`, which writes `wp-content/maintenance.php` and `wp-content/db-error.php`.

WordPress shows a fallback screen in two states, and in neither is a theme loaded:

| State | Entry point | What is alive |
| --- | --- | --- |
| `.maintenance` in the site root — `wp maintenance-mode activate`, **and core itself during every core and plugin update** | `wp_maintenance()` in `wp-settings.php`, before plugins and theme | PHP and the filesystem |
| Database unreachable | `wpdb::dead_db()` → the drop-in | PHP and the filesystem. No options, no translations |

Both `require_once` their drop-in and then `die()`, and neither sends a status header first. So the drop-in owns its own `503`, `Retry-After` and no-cache headers, and the screen it serves has to exist already — the theme prerenders it.

## Why generated, not documented

Two properties are load-bearing, and both are easy to lose in a hand-written file:

- **No dependencies.** No Composer autoloader, no WordPress function beyond the `WP_CONTENT_DIR` constant, no database. At the moment these run, that is all there is.
- **No early return.** Control goes to `die()`, not back to WordPress. A drop-in that returns when the screen is missing serves a blank page — so every path prints something, including a plain 503 sentence when the theme has not rendered its screen yet.

Both are pinned by tests, along with: `install` is idempotent, rewrites its own stale output when the theme directory changes, and refuses to touch a drop-in it did not generate; `remove` takes only its own back out.

## Verified against a real site

Installed on a local WordPress + Timber site and switched on with `wp maintenance-mode activate`. The response:

```
HTTP/2 503
retry-after: 600
cache-control: no-store, no-cache, must-revalidate, max-age=0
content-type: text/html; charset=utf-8
```

…with the theme's branded screen as the body.

**The database-outage path is covered by unit tests, not by a live check.** Killing MariaDB inside DDEV took the web container down with it, so the live attempt measured nothing and cost a `ddev restart`. The two drop-ins share one generated body, and the `.maintenance` path exercised that body end to end through Apache; what remains untested live is only WordPress's own call site, which is three lines of core.

## Rejected

**Widening the 10-minute `.maintenance` window from the drop-in.** `wp_is_maintenance_mode()` checks the expiry *before* the `enable_maintenance_mode` filter fires and *before* `wp_maintenance()` reaches the drop-in. The filter can only turn maintenance mode off, and it runs before plugins load anyway. The only lever is the timestamp inside `.maintenance`.

## Not included

Deploy wiring. `wp maintenance-mode` around the transfer belongs to `wordpress-base` (portadesign/wordpress-base#113).

## Known-failing on the base branch

`TypographyLocaleResolverTest` errors and PHPStan reports one `arguments.count` on `StarterBase.php:1344` — both from a `parisek/twig-typography` version drift, both reproduce on `main` with this branch's files removed. Untouched here.

https://claude.ai/code/session_01TWmz8DDySPpxveaxqzRESF


parisek:
## Agent review, 2026-08-11

**Codex** (`codex-rescue`, GPT-5.1-codex-max), given seven specific directions to attack rather than a general pass. Six findings, five fixed in `HEAD`.

### Fixed

| Finding | What |
| --- | --- |
| `addslashes()` corrupts a path with a double quote | It escapes `"` as `\"`, which a single-quoted PHP literal does not decode. The generated file compiled and then read a path with a stray backslash, falling through to the plain-text branch with nothing logged — because the drop-in runs where nothing can log. `var_export()` emits a literal that round-trips. |
| The write was not atomic | These files are `require_once`d by live requests, and reinstalling during an active outage is normal, so a request could parse half a file. Temp file plus rename; a short write now counts as failure. |
| Multisite served one site's screen to the network | `wp-content/` is shared while the theme is per-site. Nothing can resolve this at runtime — `db-error.php` has no database to ask — so `install` warns instead of pretending. |
| The marker was a prose attribution line | It decides whether a file may be overwritten or deleted, and "Generated by parisek/timber-kit" is exactly what somebody copies into a drop-in of their own. Now a machine-shaped token, with a test for the copied-attribution case. |
| The tests pinned structure, not behaviour | "No early return" passed on `return false;`. "No dependencies" passed on an added `get_option()`. The 503 was asserted to be *present in the source*. They now run the generated file in a bare PHP process with only `WP_CONTENT_DIR` defined and read the status code back — the environment the file actually meets. |

### Rejected

**`--screen` pointing outside the theme, reported as a path-containment gap.** Whoever runs WP-CLI already has the filesystem; this is not a privilege boundary and a containment check would only perform security. It is a real typo hazard, so a `..` segment is refused with a plain message — the cheap half of the fix, without the theatre.

### What the review cleared

Worth recording, because these were the author's own suspicions and all four were wrong:

- `status` tells the truth both after a theme rename and after a package release — it byte-compares against freshly generated source.
- The runtime needs no autoloader and no database; output buffering, a large screen file and an unreadable path are all handled.
- Trailing slashes and a relocated `wp-content` are handled; only multisite was broken.
- The quote-breakout hypothesis did not hold: `addslashes()` does neutralise `'` and `\`. Only the double-quote *corruption* variant was real — a good example of a finding that is right for a different reason than the one that prompted the look.

PR #122, opened 2026-08-12
feat(outage-screen): cover the fatal-error state with a php-error.php drop-in

### LINKED ISSUE #118: outage-screen: add a php-error.php drop-in for the fatal-error state
## Problem

`outage-screen install` generates two drop-ins: `wp-content/maintenance.php` and
`wp-content/db-error.php`. They cover a planned outage and an unreachable
database. The third failure state — a PHP fatal error — still renders the core
text "There has been a critical error on this website".

That state is the most common of the three in practice. It fires after a bad
plugin update, on sites nobody deployed to that day.

## What blocked this until now

The documented reason for leaving it alone was the recovery e-mail: WordPress
5.2+ mails the administrator a recovery-mode link on a fatal error, and a
drop-in that only renders a prettier screen looked like it would cost that path.

**That is true for `fatal-error-handler.php` and false for `php-error.php`.**
The two are separate drop-ins with separate contracts.

`wp-includes/class-wp-fatal-error-handler.php`:

- line 54 — `wp_recovery_mode()->handle_error( $error )` sends the mail
- line 60 — `display_error_template( $error, $handled )`
- line 149 — that method `require_once`s `wp-content/php-error.php`

The mail is already sent by the time the drop-in loads. A `php-error.php`
drop-in cannot reach it.

## Scope

- Generate a third file, same shape as the existing two: send headers,
  `readfile()` the prerendered screen, print a fallback on every path.
- Status `500`. No `Retry-After` — a fatal error carries no honest estimate of
  when it clears.
- Add `!php-error.php` to the nested `wp-content/.gitignore` the installer
  already documents, so the file is committed like its two siblings.

No new screen. `php-error.php` reads the same prerendered HTML the other two
read.

## Rejected: the `fatal-error-handler.php` drop-in

`wp-includes/error-protection.php:81` supports replacing the handler object, and
`display_default_error_template()` is `protected`, so subclassing
`WP_Fatal_Error_Handler` and overriding one method is a supported route that
keeps the recovery mail.

It is the wrong lever here. That drop-in registers on **every request**, not
only on failure, so a defect in it is a defect across the whole site. And it
changes handler *behaviour*, where this issue changes only the *appearance* of
the error page — which is exactly what `php-error.php` exists for.

Worth keeping in mind for a future need to change *when* or *whether* the screen
appears.

## Open questions

1. Should `install` write all three files unconditionally, or is `php-error.php`
   opt-in? The other two are inert until their trigger fires; so is this one.
2. What proves it works without breaking a site by hand? A verification must
   cover the recovery mail still arriving, not just the screen rendering. Note
   that `WP_DEBUG_DISPLAY` makes PHP print the error first, which sends headers
   and makes core skip the template (`headers_sent()` check, line 59) — a local
   test with debug display on shows a false negative.

parisek:
Closes #118.

## What

`wp timber-kit outage-screen install` writes a third drop-in,
`wp-content/php-error.php`. `WP_Fatal_Error_Handler` loads it in place of
core's "There has been a critical error on this website", and it `readfile()`s
the same prerendered screen the other two serve.

| Drop-in | Covers | Status | `Retry-After` |
|---|---|---|---|
| `maintenance.php` | deploy or core update | 503 | 600 |
| `db-error.php` | database unreachable | 503 | 600 |
| `php-error.php` | **fatal PHP error** | **500** | **none** |

## Why 500 and not 503

`503` means a planned, bounded outage, and monitoring reads it that way — some
of it suppresses alerts on a `503` carrying `Retry-After`. A crash is neither
planned nor bounded, and the one thing it must not do is look routine.
`Retry-After` is dropped for the same reason: nobody knows when a crash clears,
so the header would be a guess presented as a promise.

## The recovery e-mail was never at risk

The documented reason for leaving this state alone was the recovery-mode mail.
That applies to `fatal-error-handler.php`, not to `php-error.php`
(`class-wp-fatal-error-handler.php`: mail on line 54, template on line 60,
drop-in loaded on line 149).

## Rejected

- **Replacing `fatal-error-handler.php`.** A supported extension point
  (`error-protection.php:81`), and subclassing keeps the mail — but it
  registers on *every* request, not only on failure, and it changes handler
  behaviour where this changes only the error page.
- **One shared source for all three files.** `install()` and `status()` each
  computed it above their loop, correctly, while the files were identical. Both
  now compute per file. The `status()` case is the quiet one: a shared
  expectation reports a correct drop-in as stale, and a reinstall does not fix
  it, because it is not stale. The `install()` half is pinned by
  `test_install_writes_each_drop_in_its_own_source`, and verified by mutation.
  **The `status()` half is not covered** — the WP_CLI layer has no unit tests by
  design, so the same hoist there is caught by review only.
- **Detecting `WP_DISABLE_FATAL_ERROR_HANDLER`.** It disables the mechanism
  silently and `status` could say so. Owner's call: a site that sets it knows.
  Recorded as a known limit, not code.

## Breaking

`OutageScreen::DROP_INS` maps to `array{covers, status, retry, context}` where
it mapped to a description string. `source()` takes the filename as a third
argument, defaulting to `maintenance.php`; it throws on a name it does not
generate. Inside this package only `outage-screen status` read the old shape.
Minor version.

Projects that ran `install` before this release will see `php-error.php absent`
until they run it again.

## Also in this branch

One unrelated line: `CHANGELOG.md` carried a `||||||| parent of 0693631`
conflict marker inside the released 1.30.0 entry, added by the squash-merge of
#111. Verified that commit deleted no line, so nothing was swallowed with it.
Carried here because it is one line in a file this branch already edits.

## Checks

`composer check` green — 1134 unit + property tests, PHPStan 0 errors, ADR index
OK. The new tests read the status back out of a real subprocess run rather than
grepping the source, matching the existing suite's approach.

End-to-end verification against a real fatal error is running separately in a
consumer project; result will be posted here.

https://claude.ai/code/session_018i95RjTzeUFF9mb1i6Yt2w


parisek:
## End-to-end verification

Run against a real fatal error in a consumer project (`sloneek`, DDEV, WP 6.x,
PHP 8.3), with the branch's `OutageScreen` in `vendor/` and a temporary plugin
that fatals only on `?outage-e2e=1`. Environment restored afterwards: probe
plugin removed, `recovery_keys` and `recovery_mode_email_last_sent` deleted,
`vendor/` reinstalled from the released tag, drop-ins reverted.

| Check | Result |
|---|---|
| `install` writes three files, `status` reports three rows | pass |
| Fatal error serves the prerendered screen | pass — 444 857 B, the theme's own HTML, byte-for-byte the file `maintenance:render` wrote |
| HTTP status | `500` |
| `Retry-After` | absent |
| `Cache-Control` | `no-store, no-cache, must-revalidate, max-age=0` |
| **Recovery-mode e-mail still delivered** | **pass — 3/3 runs with the drop-in installed** |

### Two findings worth recording for anyone repeating this

**The recovery e-mail is only sent from a protected endpoint.**
`WP_Recovery_Mode::handle_error()` returns `non_protected_endpoint` otherwise
(`class-wp-recovery-mode.php:177`). A fatal on the front page therefore never
mails, with or without this change — the first attempt at verifying this looked
like a regression and was not. The check has to run against `wp-login.php` or
`wp-admin/`.

**The e-mail is rate-limited to once a day**, and the rate-limit option is
written even on an attempt whose delivery you never see. A second run inside the
window sends nothing and looks identical to a broken one; `wp option delete
recovery_mode_email_last_sent` between runs is what makes the test repeatable.

Controlled A/B, rate limit cleared each time: mail delivered **without** the
drop-in (1/1) and **with** it (3/3). That matches the code path — core mails in
`handle()` on line 54 and only reaches the template on line 60 — but it is now
measured rather than reasoned.

**`WP_DEBUG_DISPLAY` produces a false negative.** With display errors on, PHP
writes the trace first and the screen lands underneath it. Verified in this run
before switching the probe to production shape.

parisek:
## Independent review — Codex (GPT-5-Codex), 2026-08-12

Reviewed the branch diff against `main`, with a WordPress core checkout available to read.

**No findings** in four of five categories: generated-source correctness (heredoc interpolation, escaping, malformed output, wrong status), remaining shared-source assumptions across `install()`/`status()`/`remove()`, behaviour against `WP_Fatal_Error_Handler`, and factual accuracy of README/CHANGELOG.

**Three test-quality findings.** Two accepted and fixed in c665b95; one accepted only in part.

| Finding | Verdict |
|---|---|
| `remove()` never asserted anything about `php-error.php` | **Fixed.** Named explicitly. |
| The missing-screen fallback was tested for one drop-in only | **Fixed.** Runs for all three and now asserts the status alongside the printed sentence. |
| `test_the_fatal_drop_in_sets_500…` never asserted the `RETRY` marker it computed | **Half right.** The dead expression was real and is gone. The reasoning was not: an obfuscated `Retry-After` would still fail `test_install_writes_each_drop_in_its_own_source`, which asserts over the written file. It cannot be asserted at runtime either — these tests use the CLI SAPI, where `header()` is a no-op and `headers_list()` is empty. The test now states that instead of leaving the gap to be rediscovered. |

**Mutation-verified rather than assumed green.** Three deliberate breakages, each caught by at least one test: `php-error.php` given a 503 with `Retry-After`; `install()`'s source hoisted back above its loop; `php-error.php` dropped from `remove()`.

**One correction to this PR's own description.** It claimed the shared-source risk was pinned by a test. That is true of `install()` and not of `status()`, which has no unit tests by design. The description has been corrected; the `status()` half is caught by review only.

PR #124, opened 2026-08-14
feat(site-icon): emit the favicon set the theme ships, behind a flag

parisek:
## From-Project

`sloneek`, redesign instance. Removing `favicon-by-realfavicongenerator` left two favicon sets in `<head>`. The second one was not the plugin — it was `StarterBase`.

## Why

Three defects, all fleet-wide rather than project-local.

1. **`apple-touch-icon` gets an SVG.** Core asks `get_site_icon_url()` for 32, 192, 180 and 270 px; `StarterBase` returned `$favicon_path` for every one. iOS cannot read an SVG here. Today only the theme's own hardcoded PNG saves it — which is the markup this change is meant to let themes delete.
2. **Duplication is the default.** Measured across the fleet: **22 of 22** timber-kit projects with a favicon on disk hardcode the block in `layout.twig`, and **none** relies on the filter as its only source.
3. **An uploaded Site Icon is silently ignored.** `get_site_icon_url` overrides it unconditionally. The editor uploads, saves, and nothing changes.

A fourth thing is worth naming because it explains the shape of the fix: `has_site_icon()` is `(bool) get_site_icon_url( 512 )`. Filtering the getter is what opens `wp_site_icon()`'s early return. Nobody intended that; it was a side effect. It is now deliberate and commented.

## What

`$site_icon_tags`, default off per the flag doctrine in AGENTS.md. On, it filters `site_icon_meta_tags` and writes the files the theme ships.

**Discovery is by convention, not configuration** — no path property, and none added. `static/images/touch/` is probed for known filenames; a tag is written only where a file exists. That is what lets one code path serve both RealFaviconGenerator generations in circulation:

| Generation | Projects | Result |
| --- | --- | --- |
| modern (`favicon.svg`, `favicon-96x96.png`, …) | 20 | SVG + 96px PNG + ico + apple-touch + manifest |
| 2017-era (`favicon-16x16.png`, `safari-pinned-tab.svg`, …) | 6 | 32px PNG + ico + apple-touch + manifest |

A fixed path map would have emitted dead links on those six.

`theme-color` and `apple-mobile-web-app-title` come from the manifest's `theme_color` / `short_name`, so they need no configuration either — which is what lets a consuming theme drop its favicon block from the layout entirely rather than only partly.

## Rejected

- **Removing `$favicon_path`.** No project in the fleet overrides it, but it still drives the off-path, and deleting a `protected` property breaks a consumer silently instead of loudly.
- **`getimagesize()` for PNG `sizes`.** The filename already states them. Opening two files per request to learn what the name says is a poor trade.
- **A `mask-icon` tag** for the `safari-pinned-tab.svg` that 6 projects ship. It needs an author-chosen tint colour no file states; a guessed colour renders worse than no pinned-tab icon.
- **Fixing it in place instead of behind a flag.** Correct per the letter of the bug, wrong per AGENTS.md § Feature flags — it changes rendered output on upgrade.

## Downstream

Nothing changes until a project sets the flag. A follow-up PR in `tailwind-base` / `wordpress-base` drops the favicon block from `templates/layout/layout.twig`; it must land only after this is released, or the window between them leaves a site with no favicon at all.

## Verification

`composer test` 1277 green (7 new tests for the tag set, 3 for both flag branches), `composer phpstan` level 8 clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- `StarterBase::$site_icon_tags` (default off) replaces WordPress's four legacy site-icon tags with the favicon set the theme ships. Core asks `get_site_icon_url()` for 32, 192, 180 and 270 px and a theme answering with one SVG gets that SVG four times — including as `apple-touch-icon`, which iOS cannot read — plus an `msapplication-TileImage` for a Windows 8 tile. The flag probes `static/images/touch/` for known filenames and writes a tag only for a file that exists, so both RealFaviconGenerator output generations (modern `favicon.svg` + `favicon-96x96.png`, and the 2017-era 16/32 set) are handled with nothing configured; PNG `sizes` come from the filename rather than from opening the file, and `theme-color` plus `apple-mobile-web-app-title` are read from the manifest's `theme_color` and `short_name`. An uploaded Site Icon now wins over the theme's files — the off-path overrides it silently, which is why the fix is a flag and not a correction in place. `safari-pinned-tab.svg` is knowingly skipped: its `mask-icon` tag needs a tint colour no file states, and a guessed one renders worse than no pinned-tab icon.
+
 ## [1.32.0] - 2026-08-12
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -583,6 +583,50 @@ class Base extends StarterBase {
 }
 ```
 
+## Site icon tags
+
+WordPress emits four site-icon tags: `icon` at 32 and 192 px, `apple-touch-icon`,
+and `msapplication-TileImage`. It asks `get_site_icon_url()` for each size, so a
+theme that answers with one SVG gets that SVG in all four — including
+`apple-touch-icon`, which iOS cannot read, and a Windows 8 tile nobody wants.
+Themes work around it by hardcoding their own `<link rel="icon">` block in the
+layout, and then both sets render.
+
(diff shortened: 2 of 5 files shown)


parisek:
## Agent review — Codex (`codex-rescue`, 2026-08-14)

Adversarial correctness pass over the `$site_icon_tags` emitter. Two findings, both **real and both fixed** in efd6b84. Each was verified against the code before acting; neither was taken on the report alone.

| Finding | Outcome |
| --- | --- |
| An apple-only set registers both filters but answers no gate URL, so `has_site_icon()` stays false and `wp_site_icon()` returns before any tag is written | **Fixed** — registration and the gate now ask the same `favicon_image_file()`, so they cannot disagree |
| A `site_icon` option pointing at a deleted attachment is truthy but resolves to nothing, and deferring to it suppressed the whole theme set | **Fixed** — precedence keys off whether the upload actually resolves, not off the option being nonzero |

The first finding also caught a comment that lied: *"any shipped file answers here"* described the intent while the code covered three slots out of five. That mismatch is what made the bug invisible on re-reading.

**Where the fix departs from the report.** Codex proposed `reset( $found )` for the gate URL. Rejected in favour of restricting the gate to the four image slots: `reset()` would let a manifest-only theme answer its `.webmanifest` as the site icon URL, and `get_site_icon_url()` is public API that other code reads expecting a real picture — an SEO plugin's schema logo, for one. Answering nothing is the honest result there, and `test_manifest_only_set_answers_no_gate_url` pins it.

Both findings carry a regression test naming this review. `test_uploaded_site_icon_wins_over_the_theme_files` was dropped as superseded — the replacement pins *why* the upload wins, not merely that it does.

Suite: 1280 green, PHPStan level 8 clean.

PR #126, opened 2026-08-14
feat(svg): derive intrinsic dimensions so SVG stops shifting the layout

### LINKED ISSUE #125: SVG attachments ship without width/height, so every one of them shifts the layout
## Problem

An SVG attachment usually reaches the browser as an `` with **no `width` and no `height`**, so nothing reserves its box and every one of them shifts the layout when it lands.

Measured on `sloneek` (production data, 2026-08-14):

| SVG attachments | Count |
| --- | ---: |
| Total | 3520 |
| With `width`/`height` in `_wp_attachment_metadata` | 2001 (57%) |
| **Without** | **1519 (43%)** |

That is not a legacy backlog that is draining. It is growing:

| Upload year | With dimensions | Without |
| --- | ---: | ---: |
| 2022 | 1118 | 555 |
| 2023 | 551 | 348 |
| 2024 | 216 | 239 |
| 2025 | 105 | **353** |
| 2026 | 11 | **24** |

The visible cost, measured with a Chrome performance trace on the same site's homepage: a cluster of ~26 layout shifts, each naming an unsized logo SVG, contributing most of a `0.42` CLS against a `0.1` budget.

## Root cause

Two layers, and neither is a timber-kit defect.

**WordPress core produces nothing.** `getimagesize()` cannot parse SVG, so `wp_generate_attachment_metadata()` stores no dimensions for `image/svg+xml`.

**The `svg-support` plugin fills the gap only halfway.** It hooks `wp_generate_attachment_metadata` and calls `bodhi_svgs_get_dimensions()`, which reads exactly two things:

```php
$attributes = $svg->attributes();
$width  = (string) $attributes->width;
$height = (string) $attributes->height;
```

There is **no `viewBox` fallback**. An SVG exported with only a `viewBox` — the default from current Figma and Illustrator — yields `''`, and `intval('')` stores `0`. That is why the gap widens with each year: newer tooling emits `viewBox`-only more often.

Verified on the 1519 rows above: a 400-row sample was **400/400 derivable from `viewBox`**, none of them carrying `width`/`height` attributes.

## What timber-kit already does right

`Helpers::formatImageFrom()` carries the dimensions through correctly the moment they exist. Measured against two attachments of the same file, one with metadata and one without:

```
id=28803  meta=(60,26)      formatImageFrom=(60,26)
id=28804  meta=(NULL,NULL)  formatImageFrom=(NULL,NULL)
```

So the `> 1` guard at `Helpers.php:50` and `Helpers.php:102` (the core [#26256](https://core.trac.wordpress.org/ticket/26256) 1px workaround) is **not** the bug. It has nothing to read.

**This means no read-path change is needed.** The whole fix is the data layer, which keeps the change small and keeps rendered output a consequence of correct data rather than of new rendering logic.

### 1. Fill missing SVG dimensions at upload

A `wp_generate_attachment_metadata` filter that resolves `width`/`height` for `image/svg+xml` from the file: `width`/`height` attributes first (stripping units), then `viewBox`'s third and fourth values.

Two rules make it safe to run alongside `svg-support`, `Safe SVG`, or anything else in this space:

- **Priority above 10**, so it observes what other plugins wrote rather than racing them.
- **Fill only what is missing.** If `width` and `height` are already non-zero, return the metadata untouched. We close the `viewBox` gap; we never overwrite another plugin's answer.

Storage is `_wp_attachment_metadata`'s own `width`/`height` keys — the same structure core uses for raster images and `wp_get_attachment_image_src()` reads. Nothing proprietary, so every consumer benefits, not only Timber templates.

### 2. `wp timber-kit svg-dimensions` — one-off backfill

The filter above only fires on upload, so the 1519 existing rows need a sweep. Proposed shape, matching `prune-originals`:

```
wp timber-kit svg-dimensions [--dry-run] [--limit=<n>] [--force]
```

- Default: only attachments whose stored `width`/`height` are missing or `0`.
- `--force`: re-derive even where a value exists (for recovering a plugin's `0`).
- Reports derived / skipped / unreadable counts.

A deliberate sweep rather than a lazy on-render write, for the same reason `OriginalImagePruner` is one: a render should not write to the database, and a read-only replica or a cached page would make the healing arbitrary.

## Open question: flag or not

`AGENTS.md` § *Feature flags & breaking changes* says anything changing rendered output ships behind a `StarterBase` flag, default off. This does change rendered output — `` tags gain `width`/`height`.

Against that, #123 (`fix(resizer): report the dimensions the file has`) landed **without** a flag, on the argument that every changed value moves from wrong to right and `0` is not a number anyone relied on. The same argument fits here: an absent attribute is not something a layout can depend on, and the added attributes only let the browser reserve the box it was going to use anyway.

There is a real difference worth weighing, though: #123 changed values the library itself computed, while this writes to the database, where the change outlives the package version.

Suggested split, if it helps: the CLI command is opt-in by nature and needs no flag; the upload filter is the part that would carry one. Happy to go either way — say which.

## Not included

Whether an `` gaining `width`/`height` can regress an existing layout. It should not: attributes never beat CSS, and they only supply an aspect ratio before load. But that is reasoning, not measurement, and it is worth one downstream project's visual sweep before this defaults to on.

parisek:
Closes #125.

## Problem

An SVG attachment usually reaches the browser as an `` with no `width` and no `height`, so nothing reserves its box and every one of them shifts the layout as it lands.

Core measures nothing for SVG — `getimagesize()` cannot parse it. The `svg-support` plugin fills part of the gap, but `bodhi_svgs_get_dimensions()` reads only the `width`/`height` attributes on the root, with no `viewBox` fallback, so a `viewBox`-only export stores `intval( '' )`, i.e. `0`.

Measured on one production library (3520 SVGs) before writing any code:

| Upload year | With dimensions | Without |
| --- | ---: | ---: |
| 2022 | 1118 | 555 |
| 2023 | 551 | 348 |
| 2024 | 216 | 239 |
| 2025 | 105 | **353** |
| 2026 | 11 | **24** |

**1519 of 3520 unsized, and the share growing** — newer export tooling emits `viewBox`-only more often, so the plugin fails more than it used to, not less.

## What this adds

- `Parisek\TimberKit\SvgDimensions` — the resolver.
- `wp timber-kit svg-dimensions [--dry-run] [--force] [--limit] [--verbose]` — sweeps the existing backlog. Registered unconditionally; a command a human types is opt-in by being typed.
- `StarterBase::$svg_dimensions` (default **off**) — the `wp_generate_attachment_metadata` filter for new uploads.

## Designed to coexist, not compete

This was the main design constraint, since every project in the fleet already runs `svg-support`:

- **Core's own structure.** Dimensions land in `_wp_attachment_metadata`'s `width`/`height` — what `wp_get_attachment_image_src()` reads — not a private key. Every consumer benefits, not only Timber templates.
- **Fill, never overwrite.** A non-zero stored value always stands. `--force` is the single deliberate exception.
- **Priority 20.** `svg-support` and friends hook at 10; equal priority would decide the winner by plugin load order rather than by the rule above.

A stored `0` counts as missing on purpose: it is what a reader that found no attribute writes, not a size anything can lay out.

## No read-path change was needed

`Helpers::formatImageFrom()` already carries the dimensions through the moment they exist. Verified against two attachments of the same file before touching anything:

```
id=28803  meta=(60,26)      formatImageFrom=(60,26)
id=28804  meta=(NULL,NULL)  formatImageFrom=(NULL,NULL)
```

So the `> 1` guard (the core [#26256](https://core.trac.wordpress.org/ticket/26256) workaround at `Helpers.php:50` and `:102`) is **not** the defect — it had nothing to read. That keeps this change to the data layer, and keeps the rendered-output change a consequence of correct data rather than of new rendering logic.

## Two things I got wrong until the real library disproved them

Neither would have surfaced from fixtures, and both are why the resolver parses **only the root start tag**.

**Whole-document parsing lost 15 of the 1519.** Three uploads hit libxml's 10 MB `AttValue` limit on an embedded `<image href="data:image/png;base64,…">` — the parse aborts and takes the root's own `width="573" height="445"` with it, though they sit in the first hundred bytes. Four were Illustrator exports whose root reads `xmlns:x="&ns_extend;"`, an entity the DOCTYPE declares.

**My first explanation was wrong.** I attributed the first failure to document size and wrote that into the test. A 2 MB synthetic file parsed fine; reading the actual libxml error (`AttValue length too long`, line 20) showed it was one oversized attribute in a *child*. The test comment now records the measured cause, not the plausible one.

Root-only parsing fixes both, and the tag is hand-scanned rather than regex-matched because an attribute value may legitimately contain `>`. Entity references inside it are escaped to literal text rather than resolved — resolving a DOCTYPE entity is the XXE vector, and no attribute that can carry one is ever read. The XXE surface is closed by the shape rather than by a flag.

**1519 of 1519 now resolve**, in 6.3 s across the whole library.

## Rejected, deliberately

- **A lazy write during rendering.** It would have removed the need for a CLI command entirely. A page render must not write to the database, or the healing becomes a property of who happened to request an uncached page — and stops on a read-only replica. Same reasoning as `OriginalImagePruner`.
- **Overwriting another plugin's value.** See above.
- **Reading a stored `0` as an answer.**

## The flag

Off by default, per `AGENTS.md` § *Feature flags & breaking changes* — an `` that carried no dimensions starts carrying them, which is the point, but it is still a change to markup a project may have laid out around.

#123 is the counter-precedent and its argument fits: every changed value moves from wrong to right, and an absent attribute is not something a layout can depend on. I left the flag in anyway because there is a real difference — #123 changed values the library computed, while this writes to the database, where the change outlives the package version. **Say the word and I will default it on.**

The CLI command needs no flag either way; it is the half with no open question, and it is what fixes the existing backlog on projects that never flip the flag.

## Tests

Written first, red for the right reason. 36 cases: the attribute/`viewBox` precedence, unit stripping, relative-length rejection, comma and negative-offset `viewBox` forms, fractional rounding, a `>` inside an attribute, a self-closing root, a comment mentioning `<svg`, a nested `<svg>`, XXE, both real-world parse failures above, every `backfill` status, and both branches of the flag including the priority.

`RegisterMediaHooksTest::test_never_registers_on_upload_metadata_deletion` now pins `$svg_dimensions` off explicitly. `wp_generate_attachment_metadata` is no longer a unique proxy for "we never delete metadata on upload", so without the pin that assertion would silently have been testing another flag's default.

`composer test` 1318 passed, `composer phpstan` clean at level 8.

## Not verified

Whether an `` gaining `width`/`height` can regress an existing layout. It should not — attributes never beat CSS, and they only supply an aspect ratio before load — but that is reasoning, not measurement. Worth one downstream project's visual sweep before this defaults to on, which is the other argument for shipping it behind the flag first.


parisek:
## Agent review recorded

**Codex (GPT-5.1-codex-max, via `codex-rescue`), 2026-08-14.** Verdict: *not ready for fleet deployment* — 4 blockers, 4 should-fix, 0 nits, against a green `composer test` (1318) and a clean PHPStan. A useful reminder that a green suite is not evidence.

I reproduced five of its findings locally before changing anything. **All five were real**, and I found no error in the report. The one framing I would soften is finding 2's "policy shim, not measurement" for `viewBox` — it is correct as spec-reading, and the answer is to state the policy rather than drop it, which the README now does.

### Blockers — all fixed

| # | Finding | What was wrong |
| --- | --- | --- |
| 1 | `rootStartTag()` does not identify the root | It searched for the first lexical `<svg`. A `<svg` quoted inside a processing instruction, an entity declaration or a CDATA section won. Now walks the prolog by its own rules, accepts a prefixed root only when the prefix resolves to the SVG namespace, and decodes UTF-16. |
| 2 | Lengths and `viewBox` conflated | `72pt`/`25.4mm` ignored, `1e3` rejected, and `width="60" viewBox="0 0 120 52"` discarded the stated 60. Full SVG/CSS number grammar, every absolute unit converted to pixels, one axis combined with the ratio. |
| 3 | "Fill only what is missing" was false | The pair was treated atomically, so a missing height overwrote a **valid** width. Each axis is judged separately now. |
| 4 | `fromFile()` read every byte | 22 MB for 32 bytes. Bounded incremental read; the full-library sweep went 6.3 s → 1.4 s. |

Finding 3 is the one worth naming plainly: that guarantee was written in the class docblock, the README, the changelog **and** the commit message, and none of them were true.

### Should-fix — fixed

Unfiltered metadata read; `unchanged` distinguished from `failed` (`wp_update_attachment_metadata()` returns false for both); `--limit=bogus` and `--limit=-5` are now errors instead of silently meaning *all*; a failed write exits non-zero instead of printing a warning and `Success:` together; the XXE test asserted `null` **or** the answer, which made it unable to fail — it now asserts the exact value.

### Deferred, with reason

**Extracting the CLI sweep behind injectable adapters so pagination and aggregation can be tested.** The finding is fair. But no command in this repo is tested — the "WP_CLI I/O is intentionally not unit-tested" note predates this PR and covers six commands — so changing that is a repo-wide decision rather than this PR's to make. Happy to open it as its own issue.

### Found by the tests, not the review

`Helpers` also runs under `parisek/styleguide` with **no WordPress**, where `get_attached_file()` does not exist. Five existing tests failed the moment the read path was added, which is exactly what they are for.

### Scope change since the description above

`Helpers` now resolves a missing axis on the **read** path, always on, per owner decision — a template gets dimensions without anyone running the sweep or flipping the flag. `$svg_dimensions` governs only the upload filter. The stored-metadata writers stay, because the media library, `wp_get_attachment_image_src()` and srcset never come through `Helpers`.

Second owner decision: **refusing beats guessing.** Every ambiguity now returns null, because a wrong number is written to the database, outlives the package version, and the next sweep without `--force` reads it back as authoritative.

### Measured after

1519/1519 resolve. 152 of 152 SVG `` elements on the consuming site's homepage carry `width` and `height`, with no database write.

**And a result that limits what this PR can claim:** homepage CLS is **0.18 before and 0.18 after**. Sizing the images removed the tail of ~26 small shifts, but the score is dominated by a single 0.1496 shift that is not an image — the announcement bar carries `x-cloak`, so it has zero height until Alpine boots at ~304 ms and then expands 50 px, pushing the page down. That belongs to the consuming theme, not to this package. Worth knowing before anyone treats this PR as the CLS fix.

PR #127, opened 2026-08-17
feat(gtm): load the container from the kit, configured in code

parisek:
## From-Project

`eprukaz` / `oekoplan` (Asana: Oekoplan > Server Side Tracking). A tagging vendor asked for the container ID to be dropped from the GTM loader URL. Both sites run server-side GTM behind a randomly generated path, where the ID is redundant and hands blockers the pattern the random path exists to avoid.

## Why

The request was blocked by a dependency neither site uses for anything else.

- **GTM4WP 1.22 cannot do it.** `frontend.php:1297` builds the URL as one literal containing `?id='+i`, with no filter over it. Hiding the change in *Custom path* fails too: the charset check at `:1287` rejects `?` and `=` and **silently** falls back to `gtm.js`, which is worse than the status quo.
- **GTM4WP 2.0 can** — a per-container *Omit container ID* checkbox. Verified on a local DDEV copy with the production database: migration carried the container across, the switch produced an ID-free URL, no PHP errors. But 2.0 ships only as `2.0.0-beta3`, marked by its author as *not for production*.
- **The plugin is doing nothing else for these sites.** Measured on production: the whole frontend output is the snippet, a `noscript` iframe and `dataLayer_content = []`, with all ~40 data-layer options off. GTM4WP's value is its ecommerce data layer; these sites have no WooCommerce.

So the loader moves into the kit, where the ID rule is one line and the sites stop carrying a dependency for a `<script>` tag.

## What it means

```php
class Base extends StarterBase {

	protected array $gtm_containers = array(
		'default' => array( 'id' => 'GTM-XXXXXXX', 'domain' => 'windstream.example.com', 'path' => 'aBcDeF/' ),
		'de'      => array( 'id' => 'GTM-YYYYYYY' ),   // inherits domain + path
		'fr'      => '',                               // stated and blank: do not measure here
	);
}
```

```twig
{{ gtm_container() }}          {# <head>, after the consent defaults #}
{{ gtm_container_noscript() }} {# first thing inside <body> #}
```

Server-side tagging is optional — omit `domain` and `path` and you get Google's standard loader for the container ID.

- **Off by default.** Empty `$gtm_containers` → `gtm_container()` prints nothing and `gtm_container_noscript()` delegates to GTM4WP. Upgrading the kit changes no site's markup, which is what lets projects migrate one at a time behind one shared layout. The delegation sits on the `<body>` call because `gtm4wp_the_gtm_tag()` emits the plugin's `noscript` iframe and nothing else — the plugin injects its own script through `wp_head`.
- **A stated `path` omits the ID** and starts the query string at `?l=` instead of `&l=`.
- **Google's snippet verbatim.** Same line breaks, same `<!-- Google Tag Manager -->` comments, no vendor attributes (`data-cfasync`, `data-pagespeed-no-defer`) and no generator marks. Whole-output tests pin both shapes; a reader diffing the page source against Google's documentation finds exactly one difference — the id-less URL.
- **The `noscript` iframe prints only where it can be honest.** `ns.html` takes the container ID as a query parameter and has no ID-less form, so it prints by default without a custom path and stays silent with one. Per-container `'noscript' => true|false` overrides either way.
- **Language-keyed with inheritance.** Keys are WPML language codes (`cs`, `de`, `pt-br`), not locales; case and `_`/`-` fold. A regional variant resolves through its base language — `de-at` → `de` → `default`. A language **written out and left blank** is switched off instead of inheriting, which is the only way to say "do not measure here" in a model where everything else falls through.
- **`TIMBERKIT_GTM_ENABLED` gates the environment**, defaulting to production-only. Not `WP_DEBUG` — quietening a log must not switch measurement off with it. This also retires the `DEACTIVATE_PLUGINS` workaround.
- **Duplicate loaders are diagnosed, never suppressed.** The render path does not read the plugin's settings at all; a wrong guess there would stop measurement silently. The new `gtm_container_not_duplicated` Site Health check reports it instead, honouring the plugin's placement numbering and `GTM4WP_HARDCODED_GTM_ID`.

Deliberately **not** carried over: GTM environments (`gtm_auth` / `gtm_preview`) for a server-side path, which has no notion of them.

`twig_gtm4wp_the_gtm_tag()` is **deprecated, not removed**; the plugin stays supported for sites that use its data layer.

## Downstream

No sibling needs to re-sync — nothing changes until a project fills in `$gtm_containers`. First adopters would be `eprukaz` and `oekoplan`, as a separate step after this merges. `wordpress-base` would then swap `{{ gtm4wp_the_gtm_tag() }}` for the two calls in its layout, which is safe for unmigrated projects.

**The migration order matters and README documents it:** switch the plugin off *first*, deploy the configuration after. The reverse leaves both sources live for the length of the deploy, and a doubling reads as growth rather than as a fault.

## Checks

`composer test` (1439), `composer phpstan` (level 8, clean), `composer adr` green. 72 tests cover the feature: language resolution and blank-entry opt-out, both snippet shapes as whole-output comparisons, the `noscript` decision, every validation refusal, the environment gate, the Site Health check and the delegation.

Three Codex review passes, ten findings, all closed or explicitly rejected — recorded in a comment below.

ADR 0005 records the decision; README documents the class, the configuration, placement in the layout and the migration steps.

https://claude.ai/code/session_01Mpzvvj1KdXPt8ZJZCuytkk


parisek:
## Agent review record

Three review passes by **Codex (`codex-rescue`, gpt-5.3-codex), 2026-08-17**, run against the branch between commits. Every finding was verified against the code before acting on it. Summary, not raw output.

### Pass 1 — one blocker

| Finding | Outcome |
| --- | --- |
| GTM4WP placement numbering read backwards | **Fixed** (`dc815c7`) — later removed entirely, see below |

Verified in the plugin source: `GTM4WP_PLACEMENT_FOOTER = 0`, `OFF = 3`. The guard read `0` as off, which was wrong in both directions — a plugin on its own default placement went undetected and double-counted, and a plugin correctly switched to OFF suppressed the kit's loader and stopped measurement. **The existing test encoded the same wrong assumption**, so it passed against broken behaviour; it was replaced by a provider covering every placement plus the absent key.

Security of the emitted snippet and repo doctrine: no findings.

### Pass 2 — three findings

| Finding | Outcome |
| --- | --- |
| `GTM4WP_HARDCODED_GTM_ID` bypasses duplicate detection | **Fixed by removal** (`25a206d`) |
| Canonical-equivalent config keys overwrite silently | **Documented, not enforced** |
| `noscript` block missing from the installation | **Implemented** (`91e389e`) |

The first was resolved by deleting the render-time guard rather than patching it. Reading another plugin's stored settings is a guess about a schema this kit does not own, and the two ways of being wrong are not symmetric: a wrong guess that suppresses the loader stops measurement silently, while one that does not suppress it costs a doubling visible in GTM Preview and in the data. The state moved to a Site Health check, where the same coupling is harmless — and the check does honour `GTM4WP_HARDCODED_GTM_ID`.

`de-at` and `de_AT` fold to one key and the later wins. Left as configuration error; the folding rule is documented.

The `noscript` finding was accepted after being rejected once. The first answer was to document the omission, which was the wrong half of the trade: the constraint is not that the iframe is unwanted, but that `ns.html` has no ID-less form. It now prints by default only where the ID is already public, with a per-container override.

### Pass 3 — six findings, all valid, all closed (`e80aaf8`)

| Severity | Finding | Outcome |
| --- | --- | --- |
| High | Migration order left both sources live during cutover | **Fixed** — order inverted |
| Medium | Site Health reported duplication when the kit emits nothing | **Fixed** — effective state passed in |
| Medium | Deprecation text named the wrong replacement function | **Fixed** in docblock, CHANGELOG, README, ADR |
| Low | `noscript` carried environment params for a custom path | **Fixed** — follows the loader's rule |
| Low | Stale test docblock described removed suppression logic | **Fixed** |
| Low | Canonical fixture had no provenance | **Fixed** — source and retrieval date recorded |

The high finding is the one worth reading: the documented steps deployed configuration before switching the plugin off, so every visit in that window counted twice. A doubling reads as growth, so nothing complains and the report stays wrong afterwards. Switching the plugin off first inverts the failure — that window measures nothing, which is visible as a gap, bounded by the deploy, and needs no correction later.

### Reviewer accuracy

No incorrect findings across the three passes. **Caveat worth recording:** Codex could not execute PHPUnit in its sandbox (OPcache lock-file failure on a read-only filesystem), so every pass was static inspection. Each finding was reproduced locally as a failing test before being fixed.

PR #128, opened 2026-08-17
feat(editor)!: exclude gutenberg-editor.css from the classic editor by default

parisek:
> **Default flipped to ON after review** (owner decision, 3rd commit). It briefly shipped opt-in per `AGENTS.md` § Feature flags; the rationale for overriding that rule is in the commit message and the CHANGELOG. **This makes the release a minor, not a patch.**

## From project

`sloneek`, while auditing the Theme Settings options page. The registration form's consent sentence — a `wysiwyg` field — showed **Obchodními podmínkami** as plain black body text. The stored value carries a correct `<a href>` in all five languages, and the frontend renders it as a real, brand-coloured, underlined link.

## Why

`add_editor_style()` registers a stylesheet for **both** editors: the Gutenberg canvas, and the classic TinyMCE iframe through core's `mce_css` (`wp-includes/class-wp-editor.php:600`).

`gutenberg-editor.css` is written for one of them. Its rules are scoped to `.editor-styles-wrapper` — a class Gutenberg's body carries and TinyMCE's body does not. What still reaches TinyMCE is the Tailwind Preflight:

```css
a { color: inherit; text-decoration: inherit; }
ol, ul { list-style: none; }
*, ::before, ::after { margin: 0; padding: 0; }
```

Every ACF `wysiwyg` field therefore renders a link as plain body text and a bulleted list with no bullets. Measured in the iframe before the fix: `color: rgb(51,51,51)`, `text-decoration-line: none`. After: `oklch(0.5357 0.261 276.07)`, `underline`.

**Where the underline normally comes from:** the browser's UA stylesheet. Core's `wp-content.css` carries no `a` rule at all — an earlier version of this description claimed it did. Preflight simply loads last and wins.

## What changes

An `mce_css` filter drops the file from the classic editor's list, wired by `$mce_exclude_editor_styles` — **default `true`**. `enqueue_block_editor_assets()` is untouched: that path serves the block editor, the file's intended target.

**Why default on, against `AGENTS.md` § Feature flags.** That rule exists to avoid surprising a consumer, and nobody is surprised by an editor that underlines links and shows bullets — it is the state every project already expects. Opt-in would have left the defect standing everywhere until each project noticed it. `$restrict_allowed_blocks` is the standing precedent: default on, changes the editor considerably more (it decides which blocks exist), flag documented as the escape hatch rather than the switch-on. Recorded as an override rather than done quietly, so the rule and the code are not left silently disagreeing.

**The one upgrade check.** A theme may style the classic editor from inside that same file (`body#tinymce` / `body.mce-content-body`). Excluding the file switches those rules off, and both states render, so nothing warns. Grep `static/src/css/gutenberg-editor.css` for `mce-content-body`; on a match either drop the block — the native editor now does that job — or set the flag `false`. README and CHANGELOG both carry it.

Matching is on the file name after cutting query and fragment, so `/css/my-gutenberg-editor.css` and `/css/a.css?source=gutenberg-editor.css` are left alone. The filter is a no-op when `$gutenberg_editor_styles` is false.

## What is NOT claimed

The first draft said "nothing is lost, because nothing in it applied." **That is false and is retracted in the docblock, the CHANGELOG and the README.** A theme's `settings.css` imports reach the iframe too, so a compiled file also carries `:root` custom properties, base-layer element rules (`button { cursor: pointer }`, `html { overflow-anchor: none }`) and class-scoped block styles (`.wp-block-image`, `.wp-block-separator`). None of it styles the text a `wysiwyg` field holds — which is why the exclusion is safe — but a project whose imports differ should read its own compiled output rather than trust a blanket claim.

## Verified

- `composer check` green: 1447 unit tests / 2482 assertions, 16 property tests, PHPStan no errors, ADR index OK.
- 6 tests in `tests/Unit/StarterBase/MceCssTest.php`, plus both flag branches in `RegisterAdminAndEditorHooksTest`.
- All 6 CI checks green.

## Rejected

Restoring the styling rather than removing the cause — either CSS scoped to `body.mce-content-body`, or a `content_style` string appended in `tiny_mce_before_init()`. Both work; the CSS variant was verified in a browser first. Both re-implement what the UA stylesheet already does, and the `content_style` variant additionally hardcodes a fallback link colour in PHP.

## Downstream

Every project gets the fix on upgrade, with the one grep above as the only migration step. `sloneek` currently carries the CSS-side workaround in its own `gutenberg-editor.css` and drops it when it bumps.


parisek:
## Agent review — Codex (GPT-5.3 Codex, 2026-08-17)

Reviewed adversarially, with four of my own assertions handed over to be challenged. **Every finding checked out against the code; nothing was rejected.** Fixed in the follow-up commit.

| Finding | Outcome |
| --- | --- |
| Filter registered unconditionally, default-on for every consumer — violates `AGENTS.md` § Feature flags ("never on by default") | **Fixed.** Now `$mce_exclude_editor_styles`, default `false`, gated in `registerAdminAndEditorHooks()`, both branches tested |
| `str_contains()` overmatches — strips `/css/my-gutenberg-editor.css`, `gutenberg-editor.css.backup`, `/a.css?source=gutenberg-editor.css` | **Fixed.** Compares `basename()` after cutting query and fragment; test covers all three |
| "Nothing is lost, because nothing in it applied" is refuted — `settings.css` imports also deliver `:root` properties, `button { cursor: pointer }`, `html { overflow-anchor: none }`, `.wp-block-image`, `.wp-block-separator` | **Fixed.** Verified in the compiled file and **retracted in place** in both the docblock and the CHANGELOG. The conclusion survives (none of it styles `wysiwyg` text), the blanket claim does not |
| Missing README entry for a new public method — `AGENTS.md` § README coverage | **Fixed.** Row plus a pre-flight check to run before enabling the flag |
| PR body and CHANGELOG claim core's `wp-content.css` underlines links — it has no `a` rule at all | **Fixed.** The underline is the browser UA stylesheet. Corrected wherever the claim was made |
| Test comment says core appends `?wp-mce-<version>` before the filter — it is TinyMCE's client-side `cache_suffix` (`class-wp-editor.php:1107`, vs the filter at `:600`) | **Fixed.** Test kept as deliberate defensive coverage, reason corrected |

**The flag finding is the one worth reading twice**, because it is not a formality. A project can style the classic editor from inside `gutenberg-editor.css` itself, and excluding the file switches those rules off silently. `sloneek` — the project this bug was found on — has exactly such a block, so the default-on version would have disabled the fix it was written to replace. I argued *against* a flag in the first commit message, having not read `AGENTS.md` § Feature flags.

Also surfaced, outside this PR's scope: the reviewer noticed that `sloneek` has no `is_singular('post')` router branch, so a core post published there would 503 rather than merely lose a breadcrumb step. Tracked on the project side.

Not a formal approval — an agent review recorded as a comment.

PR #129, opened 2026-08-17
feat(starter-base): resolve the JS entry through the Vite manifest

parisek:
## From-Project

`sloneek`, branch `redesign`, 2026-08-17. A HubSpot form stopped rendering on `/cs/domluvit-ukazku/` right after a deploy. The server-side pair was consistent and byte-identical to local; the browser console was the only place the cause was visible:

```
SyntaxError: The requested module './script.js'
does not provide an export named 'n'
```

## Why

`enqueueThemeScript()` addresses the bundle by a fixed path, so the entry cannot carry a content hash. Cache-busting comes from `?ver=<mtime>` instead, which covers the reference in the HTML — but not the one the bundler emits **inside a lazy chunk**.

When a module is reachable from the entry graph *and* from a lazy chunk, the bundler hoists it into the entry and the chunk imports it back out:

```js
import{n as e}from"./script.js"   // no hash, no ?ver=
```

The browser then holds two cache entries for one file. The versioned one updates on every deploy; the bare one is pinned for as long as `Cache-Control` allows.

Measured on the downstream site, not reasoned about:

| | |
| --- | ---: |
| chunks importing the entry | **5 of 52** |
| `Cache-Control` on the bare URL | `max-age=31536000` |

**The failure is not limited to a missing export.** Minified export names are positions in a table, not identities. Adding one shared module with nothing to do with the broken feature moved them:

```
before   export{ha as n, $ as t}            chunk: import{n}
after    export{qs as n, ha as r, $ as t}   chunk: import{n, r}
```

`n` was **reassigned** from one function to another. So a stale entry asked for `r` fails loudly, and asked for `n` returns the **wrong binding with no error at all**.

## What this changes

`themeScriptFile()` prefers the manifest's `isEntry` record and falls back to `script.js`.

**Backwards compatible by construction.** A theme that has not rebuilt has no manifest, so this returns the unhashed name and nothing changes. It ships as a no-op everywhere and takes effect only once a build emits a manifest — which is why it can and should land before the build config.

It also refuses a record naming a file that is not on disk. A stale manifest is worse than no manifest: it would enqueue a 404 and take the whole bundle down, a louder failure than the one being fixed.

**Rejected:** keying the lookup on `src/js/script.js`. It works today and hardcodes a build-side path this class has no business knowing. The manifest carries exactly one `isEntry` record, so the first match is the answer.

## Verification

Prototyped end-to-end downstream before proposing anything here:

| | before | after |
| --- | --- | --- |
| chunk import | `./script.js` | `./script.B7fm2cuz.min.js` |
| HTML enqueue | `script.js?ver=…` | `script.B7fm2cuz.min.js?ver=…` |
| bare `dist/js/script.js` | exists, `max-age` one year | **does not exist** |
| console | `SyntaxError` | clean |
| the form | not rendered | **rendered** |

Tests: 6 cases, four of them fallbacks, because the fallback is the path every consumer takes until it rebuilds. **Mutation-tested rather than assumed green** — removing the on-disk check fails `..._named_entry_file_is_missing`; ignoring the manifest fails `..._hashed_entry_named_by_the_manifest`. Full suite: 1453 passed, 1 skipped. `composer validate --strict` and `composer normalize --dry-run` clean.

## Sibling projects that need to re-sync

None yet — this is a no-op until a consumer's build emits a manifest.

## Follow-ups, deliberately separate

1. **`parisek/styleguide`** — `styleguide.yaml`'s `iframe.js: /dist/js/script.js` is a third consumer of the fixed name. It serves the same bundle and the same chunks, so it has the identical cache defect; the package's own `cachebust` filter documents this failure class already (`Styleguide.php` ~1221) but cannot reach a URL the bundler generates.
2. **`portadesign/tailwind-base`** — `vite.config.js` gains `manifest: true` and a hashed `entryFileNames`. **Must land after both runtimes can read a manifest.**
3. **`portadesign/tailwind-base`** — track `static/vite.config.js` in `.claude/skeleton-manifest.yaml`. It is untracked today yet md5-identical across skeletons, so a downstream can fork it silently. Separate PR, separate concern.

Drupal was **not** verified — no `drupal-base` checkout was available. Its entry enqueue needs the same treatment before the build change lands, and I am not claiming what shape that takes.

https://claude.ai/code/session_01XbtSUAg85otvbG1Gt2zKVY

### DIFF
--- a/src/StarterBase.php
+++ b/src/StarterBase.php
@@ -426,7 +426,8 @@ class StarterBase extends Site {
 	protected bool $autopopulate_breadcrumb = true;
 
 	/**
-	 * How the theme JS bundle (static/dist/js/script.js) is enqueued:
+	 * How the theme JS bundle (static/dist/js/, filename resolved by
+	 * {@see self::themeScriptFile()}) is enqueued:
 	 *
 	 * - 'module' (default) — wp_enqueue_script_module(), correct for a Vite/ESM
 	 *   build.
@@ -2038,10 +2039,15 @@ class StarterBase extends Site {
 	 * A hashed entry closes it at the root: both references name the same
 	 * immutable file, so they cannot disagree.
 	 *
-	 * BACKWARDS COMPATIBLE BY CONSTRUCTION. A theme that has not rebuilt has no
-	 * manifest, so this returns `script.js` and nothing changes for it. The
-	 * hashed name only appears once the build emits one, which is why this can
-	 * ship before the build config does.
+	 * BACKWARDS COMPATIBLE BY CONSTRUCTION. A theme with no manifest at this
+	 * exact path returns `script.js` and nothing changes for it, which is why
+	 * this can ship before the build config does.
+	 *
+	 * Stated precisely, because the code is narrower than "no-op until you
+	 * rebuild": the trigger is a manifest FILE being present here, not this
+	 * build having produced it. A manifest left by some other tooling would be
+	 * honoured — though it would also have to name a file that exists in this
+	 * same directory, which is the check below.
 	 *
 	 * The manifest has exactly one `isEntry` record, so the first match is the
 	 * answer; keying on `src/js/script.js` would hardcode a build-side path this
@@ -2083,6 +2089,13 @@ class StarterBase extends Site {
 	 * Single source of truth for the front end (assets()) and the block editor
 	 * (enqueue_block_editor_assets()). Override in a subclass that needs
 	 * dependencies, async, a different handle, or per-context behaviour.
+	 *
+	 * IF YOU OVERRIDE THIS, CALL {@see self::themeScriptFile()} FOR THE FILENAME.
+	 * An override that keeps the old `'/static/dist/js/script.js'` literal still
+	 * works, and silently gives up the hashed entry — which is the cache defect
+	 * that helper exists to prevent, reintroduced for exactly the themes this
+	 * docblock invited to customise. Nothing warns; the wrong filename simply
+	 * resolves.
 	 *
 	 * @return void
 	 */
--- a/src/StarterBase.php
+++ b/src/StarterBase.php
@@ -2025,9 +2025,12 @@ class StarterBase extends Site {
 	 * looking for an entry, because a manifest can hold several and nothing
 	 * orders them.
 	 *
-	 * `protected` so a consumer that renamed its input overrides this rather
-	 * than losing the resolution — the same escape hatch both references give
-	 * by letting the caller ask for a different name.
+	 * `protected` so a consumer that renamed its input may redeclare it. That is
+	 * NOT the same shape as the references, which let the caller pass a name at
+	 * the call site; a redeclared constant is a weaker, more obscure surface.
+	 * The closer equivalent is the `timber_kit_theme_script_manifest_key` filter
+	 * applied where this is read — reachable from a plugin, not only a
+	 * subclass.
 	 *
 	 * @var string
 	 */
@@ -2108,9 +2111,13 @@ class StarterBase extends Site {
 	 * honoured — though it would also have to name a file that exists in this
 	 * same directory, which is the check below.
 	 *
-	 * The manifest has exactly one `isEntry` record, so the first match is the
-	 * answer; keying on `src/js/script.js` would hardcode a build-side path this
-	 * class has no business knowing.
+	 * An earlier version scanned for an `isEntry` record instead of asking for a
+	 * key, on the reasoning that a manifest carries exactly one and that keying
+	 * would hardcode a build-side path. Both halves were wrong: a manifest holds
+	 * one record per input and nothing orders them, and in Sage and Laravel
+	 * alike the key IS the interface. The paragraph is kept as a correction
+	 * rather than deleted, because it is the reasoning a reader is most likely
+	 * to arrive at independently.
 	 *
 	 * @param string $dir Absolute path to the built JS directory.
 	 * @return string Filename inside `$dir`, never a path.
@@ -2130,11 +2137,16 @@ class StarterBase extends Site {
 			return 'script.js';
 		}
 
-		// `static::` so a subclass can repoint it — which also means the value is
-		// no longer under this class's control. An override of `[]` would make
-		// the array offset a fatal TypeError, so a bad override degrades to the
-		// fallback instead of taking the site down.
-		$key = static::ENTRY_MANIFEST_KEY;
+		// Sage and Laravel both let the CALLER name the asset. Nothing calls this
+		// with a name, so the equivalent surface is a filter — which is also how
+		// this package exposes every other extension point. The const stays as
+		// the default and a subclass may still redeclare it.
+		//
+		// Either route puts the value outside this class's control, so it is
+		// re-checked below: an override or a filter returning `[]` would make
+		// the array offset a fatal TypeError, and a bad extension should degrade
+		// to the fallback, not take the site down.
+		$key = apply_filters( 'timber_kit_theme_script_manifest_key', static::ENTRY_MANIFEST_KEY );
 		if ( ! is_string( $key ) || '' === $key ) {
 			return 'script.js';
 		}
--- a/src/StarterBase.php
+++ b/src/StarterBase.php
@@ -2016,6 +2016,47 @@ class StarterBase extends Site {
 	}
 
 	/**
+	 * The manifest key Vite writes for this skeleton's single JS input.
+	 *
+	 * Consulted as a PREFERENCE, never as a requirement — a consumer may rename
+	 * its input, and this class should not fail because of it.
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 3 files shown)


parisek:
## Agent review — two independent reviewers, different lenses

Run locally on the branch, 2026-08-17. Neither reviewer saw the other's output.

| Reviewer | Model | Lens |
| --- | --- | --- |
| Codex CLI 0.146.0 | GPT-5-Codex | correctness and failure modes |
| Claude Code | Opus 5 | doctrine, contract, downstream impact |

**They converged on nothing.** Eight findings, zero overlap. That describes the lenses, not the quality of either report — each found a class the other never looked at, which is the argument for using two.

Every finding was **verified against the code before being acted on**, and the two structural ones were **reproduced with a probe script** first. That mattered: both were real, and one of them refuted reasoning written into this PR's own first commit.

### Correctness lens (Codex)

| Finding | Severity | Outcome |
| --- | --- | --- |
| Several `isEntry` records, CSS first — a stylesheet is returned and enqueued as a script module | High | **Fixed** (`e3b6a99`) |
| `file: "../elsewhere/other.js"` accepted — `is_file()` resolves `..`, so the method escapes the directory its contract names | Medium | **Fixed** (`e3b6a99`) |
| The fallback `script.js` is never checked for existence, so a theme with no build enqueues a 404 | Medium | **Rejected** — see below |
| An unreadable manifest passes `is_file()` and makes `file_get_contents()` warn | Low | **Fixed** (`e3b6a99`) |

**Why the third was rejected.** It is real and it is pre-existing: enqueueing `script.js` unchecked is exactly what this method did before the branch, so the branch neither introduces nor worsens it. Changing it is a separate concern, and the change would not obviously be an improvement — enqueueing a 404 is a loud failure, while enqueueing nothing is a silent one.

**The High finding refuted this PR's own reasoning.** The first commit declined to consult the manifest key, arguing "the manifest carries exactly one `isEntry` record". That is true of today's single-input config — which belongs to the **consumer**, not to this package. A second Vite input makes it false. The rejection is withdrawn; the key is now a preference, with an unconventional key still resolving when it is the only usable entry.

### Doctrine lens (Claude)

| Finding | Severity | Outcome |
| --- | --- | --- |
| A subclass overriding `enqueueThemeScript()` silently keeps the unhashed name and loses the fix | Medium | **Fixed** (`3369175`) |
| `enqueueThemeScript()` itself untested — only the helper was covered | Low | **Fixed** (`3369175`) |
| `is_file($manifest)` trusts any manifest at that path; the "no-op until you rebuild" claim is broader than the code enforces | Low | **Fixed as wording** (`3369175`) — the docblock now states the narrower truth; no provenance check added, since a foreign manifest must still name a usable file in the same directory |
| `$theme_script_strategy` docblock still described a fixed path | Cosmetic | **Fixed** (`3369175`) |

**The first is the one worth reading.** The method's own docblock invites subclasses to override it, and an override that keeps the old literal still resolves to a real file — so nothing breaks and nothing warns. It just gives up the protection, for precisely the population that was invited to customise. The warning now sits on the invitation rather than in a changelog.

**Reviewer error worth recording.** The doctrine reviewer rated the untested `enqueueThemeScript()` as Low. It is not: the filename is concatenated twice there, once into a URL and once into a filesystem path for the version stamp. Pointing `$ver` at the fallback path yields `null` — a URL with no cache buster — while every test in the original six stayed green. The new `EnqueueThemeScriptTest` catches exactly that mutation.

### Mutation testing

Every defence was checked by breaking it, because a guard nobody has watched fail is indistinguishable from no guard.

```
bare-filename check removed   -> ..._escapes_the_directory        FAILS
.js suffix check removed      -> ..._is_not_javascript            FAILS
key preference removed        -> (nothing failed)  <- see below
$ver from the unhashed path   -> ..._versions_the_same_file       FAILS
manifest ignored entirely     -> ..._hashed_entry_named_by_...    FAILS
on-disk check removed         -> ..._named_entry_file_is_missing  FAILS
```

The third row is the useful one. The CSS test passes with or without the key preference, because the `.js` check already rejects a stylesheet — so it could not observe the preference at all. A second test using **two script entries** isolates it, and now fails when the preference is removed. One test was measuring something adjacent to what it claimed to verify.

Suite: 6 tests at the first commit, **15** now. Full run 1462 passed, 1 skipped.

### Not covered by either review

- Drupal. No `drupal-base` checkout was available, so nothing is claimed about its entry enqueue.
- The two follow-up repos (`parisek/styleguide`, `portadesign/tailwind-base`) are unopened; the build change must land after both runtimes can read a manifest.

parisek:
## Second review round — after reading the prior art

Prompted by a question I should have asked myself before opening this: **has anyone solved this already?** Two have, and both contradicted the design in commit 1.

| | Sage 9 `JsonManifest::get()` | Laravel `Vite::chunk()` | this branch, commit 1 |
| --- | --- | --- | --- |
| lookup | `$manifest[$asset] ?? $asset` | `$manifest[$file]` or throw | **scan for `isEntry`** |
| keyed by | source path | source path | — |
| missing | pass through the key | **throws** | fallback |

Neither scans. **The key is the interface.** Commit 1 declined to use it, calling it "a build-side path this class has no business knowing" — in both references that path is precisely how you name the thing you want. Commit `3246a16` reverses it.

The one place this class should *not* follow Laravel is the failure mode: an exception would fatal a site whose only sin is not having rebuilt. Sage's fallback is the right model, and it is what was already here.

### Then two reviewers, again, on the rewritten shape

Same split as the first round — Codex on correctness, Claude on doctrine — and neither saw the other's output.

**They converged on one finding, and it was the important one.** Commit `3246a16` deleted the `.js`-suffix guard, arguing a keyed lookup makes a stylesheet unreachable. Both reviewers rejected that independently, and it reproduces in one line:

```
{"src/js/script.js": {"file": "style.css"}}   ->  style.css
```

The key decides which **record** is read. It says nothing about that record's `file`. Restored in `9eee2c5`.

**The test that should have caught it was the reason it was deleted.** `test_rejects_an_entry_that_is_not_javascript` wrote its stylesheet under `src/css/style.css` — a key the lookup never reads — so it passed on the key miss, not the suffix check, and kept passing after the check was removed. The mutation run did not catch it either, because the test and the code agreed with each other about the wrong thing. Now split into two tests that each mean one thing.

### Findings and outcomes

| Finding | Reviewer | Outcome |
| --- | --- | --- |
| `.js` guard deleted on a false premise; CSS reachable under the script key | **both** | Fixed `9eee2c5` |
| `test_rejects_an_entry_that_is_not_javascript` passes for the wrong reason | Claude | Fixed `9eee2c5` — split in two |
| A `protected const ENTRY_MANIFEST_KEY = []` override fatals on the array offset | Codex | Fixed `9eee2c5` |
| Commit 4's refuted reasoning still shipped in the docblock | Claude | Fixed `0dc6eb0` |
| `protected const` is not the escape hatch the references give — a subclass only, never a plugin | Claude | Fixed `0dc6eb0` — see below |
| Two test docblocks describe order-preference the rewrite removed | Claude | Fixed `0dc6eb0` |
| TOCTOU between `is_readable()` and `file_get_contents()` | Codex | **Rejected** |

**Rejected:** the TOCTOU window is microseconds, the consequence is one log line, and the alternatives are `@` (which hides real problems) or a check with the same window. Laravel does not guard it either.

**The escape-hatch finding is the one worth reading.** The docblock claimed `protected const` mirrored the references. It does not: in Sage and Laravel the caller names the asset at the call site, and nothing calls this method with a name. That could have been fixed by editing the sentence. The mechanism was fixed instead — the sentence described the right design and the code was not doing it. The key now passes through `timber_kit_theme_script_manifest_key`, matching how this package exposes every other extension point (`timber_kit_breadcrumb_items`, `timber_kit_health_checks`, `timber_kit/block_renderer/context`). Reachable from a plugin, not only a subclass.

### Mutation testing

Every defence, including the new ones:

```
.js check removed          -> ..._stylesheet_named_under_the_script_key  fails
key-type guard removed     -> ..._non_string_key_override                errors
bare-filename check removed-> ..._escapes_the_directory                  fails
on-disk check removed      -> ..._entry_file_is_missing                  fails
keyed lookup -> first record -> 5 tests fail
filter dropped, const read directly -> both filter tests fail
```

**One mutation run in this round was itself wrong.** The keyed-lookup mutation first reported clean; the shell had split the replacement string on the `:` inside `static::`, so nothing was substituted. An unrun mutation is indistinguishable from a survived one. The harness now asserts its anchor is present before writing — the same defect class as the mis-keyed test above, one layer out.

Suite: 6 tests at commit 1, **20** now. Full run 1467 passed, 1 skipped.

### Still open

- CI has never run green on this branch. Every failure so far is `codeload.github.com` returning `429` during GitHub's partial outage — four jobs died in `composer install`, and the PR-title linter died fetching its own action. Will re-run once GitHub recovers; nothing is claimed about the checks until then.
- Drupal unverified — no `drupal-base` checkout available.
- The two follow-up repos are unopened; the build change must land after both runtimes can read a manifest.

PR #130, opened 2026-08-18
fix(starter-base): stop a module entry carrying ?ver=, which instantiates it twice

parisek:
## From which project

`sloneek`, while measuring the site against contractual speed acceptance criteria. Found on `sloneek.profi.ci` on 2026-08-18, tracing a cold mobile load at 4x CPU throttling and Slow 4G.

## Why

`1.37.0` gave the theme JS entry a content-hashed filename. That fixed the **stale-bundle** half of the chunk-imports-the-entry problem its changelog describes: both references now name the same file, so the `does not provide an export named 'n'` failure mode is gone.

**The identity half survived, and this PR is that half.**

`enqueueThemeScript()` still passes a version to `wp_enqueue_script_module()`. So the HTML asks for:

```
/static/dist/js/script.B7fm2cuz.min.js?ver=1786996712
```

while a Vite chunk imports the entry back out by its own relative path:

```js
import{t as e}from"./script.B7fm2cuz.min.js"
```

A module's identity is its **resolved URL**, not its file. Two URLs are two modules.

### Measured

One `<script type="module">` tag in the HTML. No importmap, no `modulepreload`. Two requests:

| Start | URL | Size |
| ---: | --- | ---: |
| 1300 ms | `script.B7fm2cuz.min.js?ver=1786996712` | 46 kB |
| 1606 ms | `script.B7fm2cuz.min.js` | 46 kB |

Five of the site's chunks import the entry: `contact-block`, `demo-lead-panel`, `gsap`, `picture`, `team-gallery`. Any page loading one of them pays twice.

The wasted 46 kB is the smaller cost. The entry's **top-level code runs twice** - every listener, every registration, every side effect.

## What changes

`enqueueThemeScript()` passes a version only for the unhashed `script.js` fallback.

A manifest-resolved filename already carries a content hash and **is** the cache key, so the query bought nothing on that path and cost the split. The fallback has no hash and still needs a buster.

`null`, not `false` - per `wp_register_script_module()`, `false` substitutes the WordPress version and would reintroduce the split.

**The classic `defer` strategy is untouched.** It is not a module and cannot split.

## Tests

Full suite green: 1469 tests, 2513 assertions, 1 skipped.

`test_versions_the_same_file_it_enqueues` moves to the `defer` strategy - the one that now carries a version for a hashed entry. Its intent is unchanged: it still guards against reading the stamp from the wrong path. Two new tests pin the module behaviour on both sides:

- `test_module_strategy_omits_the_version_for_a_hashed_entry`
- `test_module_strategy_keeps_the_version_for_the_unhashed_fallback`

## Downstream impact

Every project on a Vite ESM build with `theme_script_strategy = 'module'` has this today. No downstream change is needed - the fix arrives with the package.

Not verified downstream yet: `sloneek` has deliberately not patched its vendor copy, so this PR is the only carrier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   site as two 46 kB requests from a single `<script type="module">` tag, at
   1300 ms and 1606 ms of a cold mobile load.
 
-  The query bought nothing there anyway. A manifest-resolved filename already
-  carries a content hash and IS the cache key. So the version is now passed
-  only for the unhashed `script.js` fallback, which still needs one. `null` is
-  used rather than `false`, because `false` substitutes the WordPress version
-  and would reintroduce the split.
+  The query bought nothing there anyway. The content hash already IS the cache
+  key. So the version is now omitted only when the filename matches the Vite
+  content-hash convention. The `script.js` fallback and an unhashed filename
+  supplied by a manifest keep their cache buster. `null` is used rather than
+  `false`, because `false` substitutes the WordPress version and would
+  reintroduce the split.
 
   **No effect on the classic `defer` strategy**, which is not a module and
   cannot split. Its version is unchanged.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   **No effect on the classic `defer` strategy**, which is not a module and
   cannot split. Its version is unchanged.
 
+  The hash test looks for the hash itself, not for `.min`. Minifying is a
+  separate Vite setting, so an unminified build still emits
+  `script.<hash>.js` — requiring `.min` there would hand it a version query
+  and reinstate the split it just removed.
+
 ## [1.37.0] - 2026-08-17
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **A module entry no longer carries `?ver=`**, so the browser stops
+  instantiating it twice.
+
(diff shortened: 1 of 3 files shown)


parisek:
## Agent review recorded

**Codex (via `codex:codex-rescue`), 2026-08-18**, adversarial review of this PR. Two follow-up commits landed from it.

### Fixed

- **The hash test was unsound.** `'script.js' !== $file` treated any manifest-resolved name as hashed, so a manifest naming `script.min.js` would have lost its cache buster. Replaced with a pattern test, plus a regression test that fails on the old code (`c454a33`).
- **That replacement had its own gap, found on review of the review.** It required `.min.js`, but minifying is a separate Vite setting from hashing — a hashed unminified entry (`script.<hash>.js`) would read as unhashed, keep its query, and reinstate the double instantiation this PR removes. `isUsableEntryFile()` accepts any `.js`, so that build does reach the check. `.min` is now optional (`d98fc83`).
- **Verified by revert-and-run**, not by inspection: the hashed-entry test goes red when the fix is reverted. The `defer`-strategy test does not catch it alone, which is why the new tests exist.

### Rejected, with reasons

- **`null` for `$version`.** Correct and documented — core suppresses the query for `null` since WP 6.5. Not changed.
- **Block editor path.** `assets()` and `enqueue_block_editor_assets()` call the same method and only `$version` changes; handle, src, deps and strategy are untouched. No regression.
- **Coverage lost by moving `test_versions_the_same_file_it_enqueues` to `defer`.** Its hashed-path assertion is intact, and the case that mattered is now covered by the new unhashed-manifest test.

### Reviewer error worth recording

Codex could not push from its sandbox (no network), so its commit was applied and pushed from the host. Its fix was also narrower than the defect: it tested a **naming convention** where the property that matters is the **hash**. That is the same class of error as the original `'script.js' !== $file` — checking one known shape instead of the invariant — one level finer.

My own first mutation test was wrong too: deleting `(?:\.min)?` produces a third regex that breaks the *minified* case, so it proved the suite was sensitive to that line rather than that the new test guards the new case. Redone against Codex's exact form, it fails exactly one test — the intended one.

Full suite: 1471 tests, 2519 assertions, 1 pre-existing skip.

PR #132, opened 2026-08-19
feat(starter-base): send the preload list as a Link header, not only as tags

parisek:
## From-Project

`sloneek`, the day after launch. Kinsta offers HTTP 103 Early Hints; the question was whether the site is ready for it. It is not, and neither is any sibling: every project sends preload hints as HTML tags only.

## Why

Core's `wp_preload_resources` filter collects preload resources and renders them in exactly one place — `<link rel="preload">` tags in `wp_head`. A tag is found once the browser parses the document, which is the thing it was waiting for. The same list as a response header arrives before the body, and every browser acts on it.

It is also the whole of an origin's part in Early Hints. PHP cannot emit an informational response, so a 103 is always synthesised by an edge from `Link:` headers it saw on a previous 200. **The header is therefore worth sending whether or not any edge does that** — and where one does, the first visitor to a URL never benefits from the 103 itself.

## What

- `StarterBase::$preload_headers` (default `true`, filter `timber_kit_preload_headers`) — hooks `send_headers` and sends the hints as `Link:`.
- `StarterBase::$preconnect_origins` (default `[]`, filter `timber_kit_preconnect_origins`) — `rel=preconnect` origins in the same header.
- `PreloadHeaders::format()` — the formatting, with no hooks and no globals.

A project that already sets `$preload_fonts` gets the header with no further configuration, because the header reads the same filter the tags do.

### Rejected

| Rejected | Why |
| --- | --- |
| A second, header-only resource list | Two lists of one thing is one list and a bug waiting for whoever edits the wrong one |
| The page's LCP image | `send_headers` fires before the query runs. It stays a `fetchpriority="high"` attribute, where the projects already put it |
| Escaping a quote or newline out of an attribute | No legitimate MIME type, media query or srcset contains one, so such a value is a mistake worth losing |
| Truncating at the byte cap | A cut link leaves an unclosed angle bracket and takes the rest of the header with it. Whole entries are dropped from the tail |

### Refusals in `format()`

A relative URL (an edge replaying it as a 103 has no document to resolve it against), an entry without `as` (fetched with no destination, then usually downloaded again by whoever needed it). Every font gets `crossorigin` whether or not the entry asked for it — a font is fetched in CORS mode regardless, so omitting it downloads the file twice.

## Tests

18 new tests, all green. Two pre-existing failures on this checkout — one PHPUnit, one PHPStan — both `Parisek\Twig\TypographyExtension`, and both reproduce on pristine `main`: the local `vendor/parisek/twig-typography` is older than the code expects.

## Downstream

Nothing to re-sync. Sites pick it up with the next `composer update`; the header then carries whatever `$preload_fonts` already declared. `$preconnect_origins` is per project and opt-in.


parisek:
## Agent review — Codex (GPT-5.1-Codex, via `codex-rescue`), 2026-08-19

Five findings. Three fixed in `cf5d025`, one rejected with reason, one partly fixed. The review also pushed me to a correction it did not itself make.

### Fixed

| # | Finding | Outcome |
| --- | --- | --- |
| 1 | Bare `crossorigin` may not activate CORS mode when a parser reads a valueless parameter as absent — the font then downloads twice | **Fixed.** Writes `crossorigin=anonymous`, or `use-credentials` when the entry asks. Verified: this is the failure the feature exists to prevent, so a parser disagreement is not a risk worth carrying for ten bytes |
| 3 | A non-array entry reached a typed parameter and would fatal, where core's own consumer skips it | **Fixed.** Verified against `wp-includes/general-template.php:3795` — core does `if ( ! is_array( $resource ) ) { continue; }`. Also drops a non-scalar attribute value rather than stringifying an array |
| 5 | URL validation accepts `https://?x` and lets control bytes through | **Fixed.** Rejects control bytes anywhere and an authority containing a path, query, fragment or userinfo start |

### The correction it prompted

Finding 2 said `send_headers` "runs after query parsing, post-6.1" and that my in-code comment was wrong. **The comment was wrong, and the reviewer's version of why is also not quite it.** In current core, `WP::main()` runs `query_posts()`, `handle_404()` and `register_globals()` *before* `send_headers()` — this used to be the other way round.

The consequence matters more than the wording: my `is_feed()` guard **worked in current versions and would silently never fire in older ones**. The guard now reads the `WP` instance the hook already passes, whose query vars are parsed in both orderings.

The rest of finding 2 — that the filter now runs twice, and that a callback reading enqueue state can answer differently at `send_headers` than at `wp_head` — is real and is now documented on the method. No code change: a `wp_preload_resources` callback that is not a pure list of assets is already a defect on its own terms.

### Rejected

**Finding 4** — suppress the header for authenticated responses, against an edge replaying a personalised hint to an anonymous visitor.

The constraint is real. The guard is the wrong place for it. Gating on `is_user_logged_in()` would hide the feature from the only person likely to be looking for it, and would not stop a personalised hint from reaching an anonymous response by any other route. What actually holds the line is what a project puts in `wp_preload_resources` — a filter for site-wide assets. Documented on `$preload_headers` and in the README instead.

Also rejected, and this one the review did not raise: now that the queried object turns out to be reachable at `send_headers`, the page's LCP image could go in the header. It stays in the markup. A template can see the layout that decides which image that is; this hook cannot, and two places choosing it from different knowledge is worse than one place choosing it late.

### Test coverage

The reviewer's gap list was largely right. Added: `crossorigin` keyword forms including an unknown value, `use-credentials`, malformed and non-scalar entries, multi-candidate `imagesrcset` with `imagesizes`, an href-less `imagesrcset` entry (dropped — a link header has no form for it), a hostless origin, a NUL inside the URL, an AJAX guard, and an empty `feed` query var.

Not added: a live `header()` call, and a stateful-callback fixture for the double run. Both need a request; the package's unit suite has none.

**28 tests on this feature, all green.** The two pre-existing failures on my checkout (`Parisek\Twig\TypographyExtension`, one PHPUnit and one PHPStan) reproduce on pristine `main` and pass in CI — a stale local vendor, not this branch.

PR #141, opened 2026-08-24
feat(warmup): a curated warmup list that lives in code, not in wp-admin

### LINKED ISSUE #140: feat(warmup): a curated warmup list that lives in code, not in wp-admin
Follow-up to #134. The scorer already has `manual` as a source at weight 800, but the only way to feed it is Breeze's own settings row. That makes the one curated signal in the model the one signal a project cannot version.

## The problem

`SignalCollector::manualKeys()` reads `breeze-preload-cache-urls` out of `breeze_get_option( 'preload_settings' )`. So a curated list:

- is not in git — it cannot be reviewed, and nothing records why a URL is on it
- is wiped by a database import, silently, and the site quietly stops warming what mattered
- has to be re-entered per environment, and differs between them by accident rather than intent

Measured on `sloneek` while adopting #134: the site's 53 category archives scored **only** through the manual list. They are in no menu, are not a homepage, and their `lastmod` is old — so with that row emptied they fell from position 27-30 to ~2864 of 2873. That project then decided the archives are not priority, which resolved it *there*, but the shape of the problem is general.

The signals the kit has are homepage, menu membership and freshness. A priority page that is none of those has no way up:

- a campaign landing page deliberately not in a menu
- a deep page carrying real search traffic
- a category archive on a project where it *is* priority
- anything correct and finished, so its `lastmod` is old

## Proposed shape

```php
protected array $breeze_warmup_urls = array(
    '/blog/',
    '/cs/blog/',
    'https://example.de/blog/',
);
```

Entries merge into `manualKeys()`, so they score at the existing `manual` weight and need no new weight, no new tier and no ordering change.

**Both forms accepted, deliberately.** A relative path is what a project wants most of the time — the same code runs on ddev, on staging and on production, and an absolute URL would be silently wrong on two of the three (rejected by `is_local_url()`, warming nothing, reporting nothing). An absolute URL is what a project needs when the target is not on `home_url()`'s host at all.

## Which is the case with domain-per-language

`isFetchableSameHostUrl()` compares against `home_url()`'s host and nothing else. Under WPML's domain-per-language negotiation every non-default language is on a different host, so an absolute URL for one of them is rejected today.

`SignalCollector` already calls `wpml_active_languages`, which carries each language's `url`, so the allowed-host set is derivable. This feature needs it, and the sitemap crawl arguably wants the same treatment — though that is a separate change and should not ride along silently.

## Open question

Whether a relative path should be expanded across active languages, or taken literally. Literal is more predictable and matches what is written; expansion saves five lines per entry on a five-language site and removes the chance of forgetting one. My inclination is literal, with expansion left to a filter, but it is worth deciding before the API exists rather than after.

## Not proposed

Any change to ordering, weights or the cap. This adds an input to a source that already exists.

https://claude.ai/code/session_01JMYy6JHLf4mU4H4Hd47spb

parisek:
Closes #140. Follow-up to #134.

`manual` has been a scoring source since #134, but its only input was Breeze's settings row — so the one curated signal in the model was the one a project could not version. It could not be reviewed, carried no reason for any entry, differed per environment by accident, and a database import wiped it silently, after which the site stopped warming what mattered and said nothing.

`$breeze_warmup_urls` feeds the same signal from the theme:

```php
protected array $breeze_warmup_urls = array(
    '/blog/',
    '/cs/blog/',
    'https://example.de/blog/',
);
```

It adds an **input, not a tier** — ordering, weights and the cap are untouched.

## Both forms, and why that is not just convenience

An entry naming a post or page becomes its ID and comes back as `get_permalink()`. So the same written line is correct on ddev, on staging and in production without being rewritten — and under WPML's domain-per-language negotiation it comes back on that language's own host, which is why an absolute form has to be accepted at all.

**An entry that resolves to nothing is dropped.** A deleted page stops being warmed without anyone remembering the list exists, and a typo warms a 404 zero times instead of on every cycle.

`url_to_postid()` cannot see term archives, and a category archive was the motivating example, so those are kept as URLs and checked with one `HEAD` during the cron refresh — off the purge path deliberately, since that filter runs while an editor waits for a save. A failed probe **keeps** the entry: one flaky lookup must not empty a curated list, which is the stale-beats-nothing rule `runRefresh()` already follows.

## Two pre-existing bugs found while writing it

**`Helpers::extract_slug_from_url()` could fatal.** It calls `array_keys()` on the result of `wpml_active_languages` with no type check. The plugin being active does not guarantee an array — WPML answers `null` before it has booted — and `array_keys( null )` is a TypeError on a page that only wanted to resolve a link. Found by a test, not by reading.

**`Breadcrumb` dropped language-prefixed links.** It called `url_to_postid()` directly, without the fallback `formatLink()` has always had. On a WPML site with language prefixes that returns 0 for a valid `/cs/…` URL, so those global links vanished from the trail with no error.

Both call sites now use the new `Helpers::urlToPostId()`, which is the shared form of what `formatLink()` was already doing privately. Three call sites of the same idea, one function.

## The architecture test earned its keep

The first draft named Breeze in `Helpers.php`'s docblock and #137's boundary test refused the build. The helper is generic and naming a consumer in it was wrong regardless — the test caught a design slip, not a formatting one.

## Verification

1641 tests, 2722 assertions, PHPStan clean. 14 new tests cover both resolution paths, the reachability probe's three outcomes, the WPML prefix fallback and the scheme-relative refusal.

## Open question from the issue, still open

Whether a relative path should expand across active languages or stay literal. This ships **literal** — it is what is written and what can be reviewed — and expansion stays available to a project as a loop. Worth settling before the API has consumers.

https://claude.ai/code/session_01JMYy6JHLf4mU4H4Hd47spb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   relative path is correct on every environment and, under domain-per-language,
   on that language's own host. An entry that resolves to nothing is dropped, and
   what `url_to_postid()` cannot see is verified with one `HEAD` during the cron
-  refresh — never on the purge path.
+  refresh — never on the purge path. Filterable as
+  `timberkit_warmup_curated_urls`, its cap as
+  `timberkit_warmup_curated_max_entries`.
 - `Helpers::isSiteUrl()` / `Helpers::siteHosts()` — whether a URL belongs to this
   site, matching host **and** port, and counting each WPML language's own host
   under domain-per-language negotiation.
@@ -43,6 +45,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   not know yet and for a site serving its sitemap from a non-default path. A
   non-string return, or a URL on another host, is ignored in favour of the
   detected path.
+
+### Changed
+
+- The warmup refresh no longer returns early when the sitemap is empty or
+  unreachable. Curated entries are added first, so a project that names its
+  pages explicitly still gets them warmed in exactly the situation where the
+  sitemap cannot help. On a site with no curated entries this changes nothing;
+  on one with them, a refresh that used to no-op now does work.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- `Helpers::formatLink()` no longer replaces a valid link with an empty string.
+  `get_permalink()` answers `false` for a trashed post or a stale WPML
+  translation id, and the concatenations after it coerced that to `''` — so a
+  working URL silently became `''` or a bare `?query`.
 - `Helpers::extract_slug_from_url()` no longer fatals when `wpml_active_languages`
   answers with `null` while the plugin is active — `array_keys( null )` is a
   TypeError, and WPML returns null before it has finished booting.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   on that language's own host. An entry that resolves to nothing is dropped, and
   what `url_to_postid()` cannot see is verified with one `HEAD` during the cron
   refresh — never on the purge path.
+- `Helpers::isSiteUrl()` / `Helpers::siteHosts()` — whether a URL belongs to this
+  site, matching host **and** port, and counting each WPML language's own host
+  under domain-per-language negotiation.
 - `Helpers::urlToPostId()` — URL to post ID with the WPML prefix fallback and
   `wpml_object_id` translation. `formatLink()` had this logic privately;
   `Breadcrumb` and the new warmup list needed the same, so it is one function now
@@ -49,6 +52,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 - `Breadcrumb` now resolves language-prefixed URLs. It called `url_to_postid()`
   directly, which returns 0 for a valid `/cs/…` URL, so those global links were
   silently dropped from the trail.
+- The WPML prefix fallback no longer lets a foreign URL borrow a local path.
+  It compares paths, so `https://someone-else.test/about/` matched this site's
+  own About page and was rewritten to a local permalink — in `formatLink()`,
+  where the fallback has lived for years, and in `Breadcrumb`.
 - The same-host guard compared the hostname and ignored the **port**, so a
   sitemap index entry — or a filter callback — pointing at
   `https://<own-host>:9200/` passed it and was fetched. An internal service
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Added
 
+- `$breeze_warmup_urls` — a curated warmup list that lives in the theme instead
+  of in Breeze's settings row. Merges into the existing `manual` signal at the
+  weight it already carries; no new tier, no change to ordering or the cap.
+  Entries may be relative paths or absolute URLs, and both are resolved rather
+  than trusted: one naming a post or page comes back as `get_permalink()`, so a
+  relative path is correct on every environment and, under domain-per-language,
+  on that language's own host. An entry that resolves to nothing is dropped, and
+  what `url_to_postid()` cannot see is verified with one `HEAD` during the cron
+  refresh — never on the purge path.
+- `Helpers::urlToPostId()` — URL to post ID with the WPML prefix fallback and
+  `wpml_object_id` translation. `formatLink()` had this logic privately;
+  `Breadcrumb` and the new warmup list needed the same, so it is one function now
+  and both call sites use it.
 - `Breeze\WarmupSitemap` recognises **Yoast SEO** and asks for
   `/sitemap_index.xml`. Providers are now a list checked in order — AIOSEO,
   Yoast, core — so a project gets the right sitemap from whichever plugin it
@@ -30,6 +43,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- `Helpers::extract_slug_from_url()` no longer fatals when `wpml_active_languages`
+  answers with `null` while the plugin is active — `array_keys( null )` is a
+  TypeError, and WPML returns null before it has finished booting.
+- `Breadcrumb` now resolves language-prefixed URLs. It called `url_to_postid()`
+  directly, which returns 0 for a valid `/cs/…` URL, so those global links were
+  silently dropped from the trail.
 - The same-host guard compared the hostname and ignored the **port**, so a
   sitemap index entry — or a filter callback — pointing at
   `https://<own-host>:9200/` passed it and was fetched. An internal service
--- a/README.md
+++ b/README.md
@@ -708,6 +708,49 @@ fresh. The signal is free — it is already in the XML we parse — and
 resolving real publication dates would mean matching URLs back to posts
 through WPML and custom permalinks.
 
+### Naming pages yourself
+
+The scorer treats Breeze's own manual list as a strong signal, but that list
+lives in a `wp_options` row: it cannot be reviewed, carries no reason for any
+entry, differs per environment by accident, and a database import wipes it —
+after which the site quietly stops warming what mattered.
+
+`$breeze_warmup_urls` is the same signal, kept in the theme:
+
+```php
+protected array $breeze_warmup_urls = array(
+    '/blog/',
+    '/cs/blog/',
+    'https://example.de/blog/',
+);
+```
+
+It merges into `manual` at the weight that source already carries. It adds an
+input, not a tier: ordering, weights and the cap are untouched.
+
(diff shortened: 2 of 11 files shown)


parisek:
## Review — Codex (gpt-5.3-codex), 2026-08-24

Ten findings, verified against the code one by one rather than applied as a work order. Seven fixed in 276463c, three rejected or deferred with reasons below.

**The root cause it named was right**, and it explains why the findings cluster: `Helpers::urlToPostId()` resolved by **path** before checking the host, so `https://someone-else.test/about/` matched this site's own About page and came back as a local permalink. That is a **pre-existing bug in `formatLink()`**, where the fallback has lived for years — the PR only spread it to two more call sites by extracting the helper.

### Fixed

| Finding | What it was |
| --- | --- |
| Host checked after path resolution | The gate now sits **inside the fallback branch**, not in front of the function. The first lookup compares full URLs, where a foreign host can never match — so the common path never needs to know the site's hosts, which also kept 40 existing tests from having to mock `home_url()`. |
| SSRF: scheme/port/user-info ignored | `isSiteUrl()` matches host **and** port, refuses non-http(s) and refuses user-info. `example.test:8080` is a different service. |
| `sslverify => false` | Removed. A probe with verification off can be answered by anyone on the path. |
| Probe hits already-resolved posts | `resolve()` reports provenance. A post ID settles existence, so it is not probed — it cost a request per entry per refresh and dropped pages answering 405 to HEAD while serving GET. |
| **Curated URLs were only a flag** | The serious one. `enrichRecords()` marked `manual` on records the sitemap had already produced, so an entry the sitemap ranks badly or omits did nothing at all — the whole case the feature exists for. They become records now, with a null `lastmod`: an invented date buys freshness it has not earned. |
| Empty sitemap discarded the curated list | No early return before enrichment. A project listing URLs explicitly should get them warmed when the sitemap is broken — which is when it matters most. |
| `extract_slug_from_url()` early return | Skipped final normalisation, breaking its documented "always starts with `/`" contract. Falls through now. |
| Breadcrumb inherits the host bug | Same root cause, fixed by the same gate. |

### Rejected

**`urlToPostId()` translates to the current language.** Flagged as wrong for an explicit-language URL during cron. It is the behaviour `formatLink()` and `Breadcrumb` have always had, and what a caller rendering a page wants. Changing it is a behaviour change deserving its own commit and its own evidence, not a rider on a security fix.

### Reviewer's own error, for calibration

The report located the language-translation issue at `src/Helpers.php:1888-1893` and described it as introduced here. Those lines are the extracted helper; the behaviour predates this PR in `formatLink()`. The finding is real as a question, but its framing as a regression is not — worth knowing when weighing the rest of the report.

### Verification

1648 tests, 2735 assertions, PHPStan clean. 11 new tests cover the foreign-path borrow, the port, the probe scope and all four record-append cases.

Only one reviewer so far. Per `AGENTS.md` a second, independent lens would be worth it before merge — agreement between reviewers that cannot see each other is the strongest evidence a finding is real, and the curated-records bug is the kind a doctrine-focused pass might have missed while a correctness pass caught it.

parisek:
## Reviews 2 and 3 — doctrine lens and silent-failure lens, 2026-08-26

Three reviewers now, each given a different lens so agreement would mean something. Fixed in 2a55318 and e663d71.

### Where they converged: nowhere, and that describes the lenses

No finding was reported by more than one reviewer. Per `AGENTS.md` that is worth stating rather than glossing: zero convergence across three passes says the lenses were genuinely independent, not that the findings are weak. Each pass found real defects the others walked past.

The clearest illustration is `filterReachable()`, which two reviewers reached from opposite sides. Codex said *don't probe entries a post ID already proved*. The silent-failure pass said *the accept range is wrong for the ones you do probe*. Both were right, neither implied the other.

### Review 2 — doctrine and downstream impact

**No filter, breaking the pattern every sibling knob follows.** `$breeze_warmup_priority_weights` has `timberkit_warmup_priority_weights`; the sitemap cap has `timberkit_warmup_sitemap_max_urls`; the module toggle has `timberkit_warmup_sitemap_enabled`. The curated list had none, and its 200-entry cap was a hardcoded constant — leaving it the one warmup setting a project keeping config in an mu-plugin could not reach. Now `timberkit_warmup_curated_urls` and `timberkit_warmup_curated_max_entries`. Unlike the weights filter it need not be pure, since nothing fingerprints it; the cost is a delay to the next refresh, and the docblock says so.

**No README section**, while every sibling flag has one and the CHANGELOG treats this as the headline addition. Added, mirroring how the weights are documented.

Also added the behavioural floor change it flagged as undocumented: the refresh no longer returns early on an empty sitemap.

### Review 3 — silent failure

**The probe dropped live pages.** It kept 2xx/3xx and dropped everything else, so 405, 401, 403 and 5xx removed a curated entry permanently and re-removed it every cycle, silently.

405 is the one that stings, and the reviewer named exactly why: `resolve()`'s own comment says a probe would *"drop a page that answers 405 to HEAD while serving GET perfectly well"* — and that reasoning was applied to the post-ID path only, leaving the branch the probe exists for wide open. **The comment gave false confidence the case was handled everywhere.** A term archive behind a WAF that refuses HEAD is a live page.

It drops on 404 and 410 alone now. Authorisation says nothing about existence and a warmer is not logged in; a server erroring today is not a page deleted, and dropping on 5xx would empty a curated list during an outage — when nothing should change.

**Probing was unbounded in wall-clock terms.** 200 entries × 5s serially. A `max_execution_time` fatal is not a `Throwable`, so it walks past `runRefresh()`'s catch *and* its `finally`; the refresh vanishes mid-run and the lock is left to its TTL. `PROBE_BUDGET` caps the total at 20s and keeps what it did not reach.

**`formatLink()` could blank a good link.** `get_permalink()` answers `false` for a trashed post or a stale WPML translation id, and the concatenations after it coerce that to `''`. Pre-existing — but this PR rewrote those exact lines and had the identical guard two files away in `CuratedUrls::resolve()`. Backported.

### Rejected

**An ADR** (review 2, its own confidence 55). `docs/adr/0006` sets a precedent, but the calls here — where the host gate sits, probe at refresh never at purge, a failed probe keeps the entry — are each three lines from the code they govern. An ADR would move that reasoning further from the code, not closer.

**Translating to the current language** (review 1). Unchanged behaviour that `formatLink()` and `Breadcrumb` have always had; changing it deserves its own commit and its own evidence.

### Confirmed correct, worth recording

Review 3 explicitly checked and cleared three things: `isSiteUrl()` fails **closed** on an empty host set, which is the safe direction; the `extract_slug_from_url()` guard preserves its leading-slash contract instead of short-circuiting; and the silent drops in `keys()` are the right place to spend this package's no-logging budget, since a curated list's feedback loop is code review rather than telemetry. Review 2 cleared the `src/Breeze/` boundary question — `CuratedUrls` depending on core `Helpers` is the enforced direction, and `Helpers` gained no Breeze knowledge.

### Verification

1658 tests, 2745 assertions, PHPStan clean. Eight of the new tests are table-driven over status codes, because the gap review 3 found was that only 404 had ever been exercised.

parisek:
**Open question from #140 is now decided: entries are literal.** Not an omission — see [the reasoning](https://github.com/parisek/timber-kit/issues/140#issuecomment-5423234617). In short: WPML translates slugs rather than prefixing them (`/pricing/` → `/it/listino-prezzi/`), so prefix expansion would emit 404s that the probe drops silently. If expansion is added later it must go through the post ID, and it would be an addition rather than a change to how existing entries behave — so shipping literal costs no future option.

parisek:
## Measurement: what a curated entry actually scores

Recorded here because it decided a design question on #138, and because the docblock on this branch stated it wrongly (corrected in the commit above).

Scored with the real `Scorer`, default weights:

| Record | Default | With `types['page'] = 400` |
| --- | --- | --- |
| **Curated entry, not in the sitemap, not in a menu** | **800** | **800** |
| Menu page edited yesterday | 800 — tie | 1200 — wins |
| Menu page edited last week | 700 | 1100 — wins |
| Plain page edited yesterday | 300 | 700 |
| Front page | 1800 | 1800 |

A curated entry is **competitive** with the sitemap, not ahead of it. Its 800 is flat: `manual` plus whatever `menu`/`front_page` enrichment finds, and the two signals that would raise it — freshness and type — are exactly the two an appended record does not carry.

## One correction to my own earlier reading

`appendMissingManual()` runs *inside* `enrichRecords()`, before the enrichment loop, so appended records **do** get `menu`, `front_page` and `lang`. Only `lastmod` and `type` are missing. The two are not the same kind of gap: `lastmod` is genuinely unknown and null is the honest answer, while the post type is knowable for anything that resolved to an ID — `resolve()` has `$post_id` in hand and discards it. Filed as #148 rather than fixed here, because it changes three signatures and this branch is ready.

## What it settled on #138

The tail drain must **not** exclude curated URLs. Measured against `TailPlanner::split()` directly:

| Scenario | kept | tail | curated in tail | **duplicates** |
| --- | --- | --- | --- | --- |
| 30 curated, 3000 sitemap, cap 200 | 200 | 2830 | 0 | **0** |
| 200 curated (cap exactly full) | 200 | 3000 | 0 | **0** |
| 500 curated | 200 | 3300 | 300 | **0** |
| no curated | 200 | 2800 | 0 | **0** |

`split()` subtracts everything already kept, so a URL can never be in both lists — double warming is impossible by construction, not by luck. The only case an exclusion would change is a curated entry pushed past the cap, and there the exclusion would strand it: out of the cap because it did not fit, out of the tail because it is curated. The pages a project named as important would be the only ones left cold.

The scoring above is why that case is ordinary rather than rare.

PR #143, opened 2026-08-26
feat(warmup): resolve the sitemap per SEO plugin, Yoast included

### LINKED ISSUE #142: WarmupSitemap: Yoast sites resolve no sitemap, and the failure is silent
## Summary

`WarmupSitemap` resolves the sitemap root from two candidates only — AIOSEO and WordPress core. A site running Yoast SEO gets neither, and the module produces an empty URL list without any error.

The flag is on, the cron job runs, the store stays empty. Nothing in a log or an admin screen says why.

## Where

`src/Breeze/WarmupSitemap.php` (v1.39.0):

- `resolveSitemapRootUrl()` (L541) picks `/sitemap.xml` when `isAioseoActive()` (L556), `/wp-sitemap.xml` otherwise (L546). The method is `private` and carries no filter, so a project cannot point it elsewhere.
- `fetchBody()` (L709) sends `'redirection' => 0` (L719).

## Why the core fallback does not recover it

Yoast redirects `/wp-sitemap.xml` to its own `/sitemap_index.xml` with a 301. `fetchBody()` does not follow redirects, so the response code lands outside 200–299, the body is discarded, and `fetchAndParseSitemap()` returns an empty array.

Measured on a live Yoast Premium site (WordPress 6.x, WPML, five languages), calling `fetchAndParseSitemap()` by reflection:

| Input | Records |
| --- | --- |
| `/wp-sitemap.xml` — what the code requests | 0 |
| `/sitemap_index.xml` — what Yoast serves | 1113 |

The data is fine and the sitemap is reachable from the server itself: `wp_remote_get( home_url( '/sitemap_index.xml' ) )` answers 200 with a valid `<sitemapindex>`, and one child urlset alone parses to 368 records. Only the address is wrong.

## Why it is worth a fix rather than a note

The failure is silent by construction. `runRefresh()` swallows every `Throwable` by contract, and an empty `fetchSitemapRecords()` is a normal return value, not an error. A project switches `$breeze_warmup_sitemap` on, sees no warning, and ships a warmup that preloads nothing.

Yoast is not an edge case in this fleet.

## Proposed change

1. Add a Yoast branch to `resolveSitemapRootUrl()` — detect with `defined( 'WPSEO_VERSION' )` and return `/sitemap_index.xml`.
2. Add a filter, `timberkit_warmup_sitemap_url`, over the resolved root. A private method with a fixed candidate list has no escape hatch, and the next SEO plugin will need one again.

**Keep `'redirection' => 0`.** Following redirects on a URL the module builds itself is the SSRF surface this module deliberately closed; the defect is the missing branch, not the hardening. A fix that opens redirects instead of naming Yoast trades a real protection for a symptom.

## Ordering note

Detection order matters if both plugins are somehow present: AIOSEO and Yoast each claim the sitemap, and whichever is checked first wins. Worth an explicit comment on the chosen order rather than leaving it implied by statement order.

parisek:
Closes #142.

## What

`Breeze\WarmupSitemap` resolved the sitemap root from two candidates, AIOSEO and WordPress core. A Yoast site got core's and produced an **empty** warmup list without a word of complaint.

Providers are now a list, checked in order:

| Provider | Detected by | Root |
| --- | --- | --- |
| AIOSEO | `aioseo()` / `\AIOSEO\Plugin\AIOSEO` | `/sitemap.xml` |
| Yoast SEO | `WPSEO_VERSION` / `\WPSEO_Sitemaps` | `/sitemap_index.xml` |
| WordPress core | nothing else answered | `/wp-sitemap.xml` |

Plus `timberkit_warmup_sitemap_url`, which receives the resolved URL and the detected provider key.

## Why the core fallback did not already cover it

Yoast redirects `/wp-sitemap.xml` to its own index with a 301, and every fetch here sends `redirection => 0` deliberately — following a redirect is the SSRF surface this module closes. So on a Yoast site the core path does not degrade to a slower answer, it degrades to **no** answer: the code lands outside 200–299 and the body is dropped.

Nothing reports that. An empty `fetchSitemapRecords()` is a normal return value, and `runRefresh()` swallows throwables by contract. A project switches the flag on, sees no warning, and ships a warmup that preloads nothing.

## Measured, not assumed

Against a live Yoast Premium site (WPML, five languages), by reflection:

| | Released v1.39.0 | This branch |
| --- | --- | --- |
| Provider detected | — | `yoast` |
| Root requested | `/wp-sitemap.xml` | `/sitemap_index.xml` |
| Records | **0** | **2873** |
| `breeze_preload_urls` | 0 | 200 (the cap), five language homepages first |

Post types came out of Yoast's filenames unchanged: `blog=1469 lexicon=767 page=368 case_studies=103 ebooks=61`.

The two patched files were copied onto that server for the measurement and restored afterwards; MD5 confirms the vendor tree is byte-identical to v1.39.0 again.

## What was deliberately left alone

**`'redirection' => 0` stays.** Opening redirects would also make Yoast work, and would trade a real protection for a symptom. The defect is a missing branch, not the hardening.

**`SourceNaming` logic is unchanged.** Yoast shares AIOSEO's `<type>-sitemap.xml` shape, and the pattern already tolerates the index Yoast appends when a type spills over one file (`blog-sitemap2.xml`). Only the constant was renamed — `AIOSEO_NON_POST_TYPE_INDEXES` → `NON_POST_TYPE_INDEXES` — because a list that now serves both plugins cannot keep a name claiming one. `author` comes from both; the rest only from AIOSEO.

**Detection does not read the active-plugin list.** A must-use load, a renamed directory or a bundled copy all keep the plugin's symbol and lose the list entry.

## Ordering

AIOSEO stays first, so a site that already ran it resolves exactly the path it resolved before this map existed. A regression test pins that, and it passes both before and after the change on purpose.

## Tests

Five added to `FetchSitemapUrlsTest`. Three fail on `main` (Yoast root, filter override, filter arguments); two are guards that pass on both sides by design — the AIOSEO-wins-over-Yoast ordering, and the off-host filter rejection, which `main` passes only because it has no filter at all. Saying so rather than counting it as proof.

`composer check` green: 1632 tests / 2714 assertions, 18 property tests, PHPStan clean, ADR index OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   Yoast, core — so a project gets the right sitemap from whichever plugin it
   runs, and the order is stated rather than implied. AIOSEO stays first, so a
   site that already had it resolves exactly the path it resolved before.
+- A Site Health check, `warmup_sitemap_resolved`, registered only when
+  `$breeze_warmup_sitemap` is on. The module degrades silently by design — an
+  empty sitemap result is a normal return value and the refresh job swallows
+  throwables by contract — which is right for the purge path and wrong for the
+  administrator, who was left with a warmup that preloads nothing and says
+  nothing. The check reports the one fact worth acting on: the feature is on
+  and the list is empty. A store that has never been refreshed is reported as
+  good, because a fresh install waits for its first purge through no fault of
+  its own.
 - `timberkit_warmup_sitemap_url` filter, receiving the resolved URL and the
   detected provider key. It is the escape hatch for a provider the list does
   not know yet and for a site serving its sitemap from a non-default path. A
@@ -21,6 +30,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- The same-host guard compared the hostname and ignored the **port**, so a
+  sitemap index entry — or a filter callback — pointing at
+  `https://<own-host>:9200/` passed it and was fetched. An internal service
+  bound to another port on the site's own host was reachable that way. Ports
+  are now compared, normalised by scheme first so one origin written two ways
+  still matches. Pre-existing; the sitemap-index path carried it before this
+  branch added a second way to reach the guard.
 - A Yoast site produced an **empty** warmup list, silently. Yoast redirects
   `/wp-sitemap.xml` to its own index with a 301 and every fetch sends
   `redirection => 0` (the SSRF guard, deliberately kept), so the core fallback
@@ -29,6 +45,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   because an empty result is a normal return value and `runRefresh()` swallows
   throwables by contract. Measured on a live Yoast Premium site: 0 records from
   the requested path, 1113 from the one Yoast serves.
+- Yoast is detected on its sitemap switch, not on its symbol alone. Yoast can
+  be loaded with its XML sitemap turned off, and core then serves
+  `/wp-sitemap.xml` again — detecting on the symbol would have sent exactly
+  those sites to an address that 404s, breaking a configuration that worked
+  before Yoast was recognised here.
 
 ## [1.39.0] - 2026-08-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- `Breeze\WarmupSitemap` recognises **Yoast SEO** and asks for
+  `/sitemap_index.xml`. Providers are now a list checked in order — AIOSEO,
+  Yoast, core — so a project gets the right sitemap from whichever plugin it
+  runs, and the order is stated rather than implied. AIOSEO stays first, so a
+  site that already had it resolves exactly the path it resolved before.
+- `timberkit_warmup_sitemap_url` filter, receiving the resolved URL and the
+  detected provider key. It is the escape hatch for a provider the list does
+  not know yet and for a site serving its sitemap from a non-default path. A
+  non-string return, or a URL on another host, is ignored in favour of the
+  detected path.
+
+### Fixed
+
+- A Yoast site produced an **empty** warmup list, silently. Yoast redirects
+  `/wp-sitemap.xml` to its own index with a 301 and every fetch sends
+  `redirection => 0` (the SSRF guard, deliberately kept), so the core fallback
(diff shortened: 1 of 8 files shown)


parisek:
## Review — Codex (GPT-5.3-Codex, 2026-08-26)

Verdict was request-changes on six findings. Each was checked against the code before acting; two changed the branch, one changed it more than the reviewer's severity suggested, and two were rejected with reasons.

### Fixed

**Same-host guard ignored the port** (reported Major). Confirmed: `isFetchableSameHostUrl()` compared `host` and nothing else, so `https://<own-host>:9200/` passed and was fetched. Ports are now compared, normalised by scheme so one origin written two ways still matches.

One correction to the report's framing: it presents this as a gap the filter opens. The guard is pre-existing and already governed every sitemap-index `<loc>` entry — which is the path an **attacker** controls, whereas the filter is developer-written code. The finding is right and the fix is worth having; the branch did not introduce it. Both paths now have a test.

**Yoast detected on its symbol alone** (reported Minor — this was the most serious finding on the list). Yoast carries a switch that turns its XML sitemap off, and core then serves `/wp-sitemap.xml` again. Such a site answers on the core path and 404s on `/sitemap_index.xml` — it worked before this branch and would have stopped after it. That is a regression the branch introduced, not a hardening opportunity. Detection now also reads Yoast's own switch, defaulting to on when the option is absent, which is Yoast's default.

**Empty warmup list reported nothing** (reported Major). Agreed on the problem, not on logging as the remedy: the silence is a deliberate contract on a hot path, and `runRefresh()` swallows throwables so a sitemap outage can never surface as a fatal in cron. Added a Site Health check instead, `warmup_sitemap_resolved`, registered only when the feature is on. A never-refreshed store reports good — a fresh install waits for its first purge through no fault of its own, and a check that cries wolf there teaches people to ignore it.

### Not adopted

**Fall through to the next provider when the detected one returns nothing.** The Yoast-switch fix covers the configuration that actually produces this. A blind second fetch on every failed refresh would mask real misconfiguration and pay a network round trip to do it.

**Detection should prove the sitemap endpoint exists, not just the symbol.** That is a probe on the refresh path for a condition the switch already answers. Same reasoning.

### Test-quality finding: accurate

The report correctly identified which added tests pass on `main` too. That matches what the PR description already said. Five more tests are added here; three fail without this change (both port cases, the Yoast switch), and the two that pass on both sides are guards, labelled as such.

### Reviewer accuracy

No fabricated line references; the two file/line ranges checked pointed at the right code. The one thing to carry forward is the severity inversion — the finding filed as Minor was the regression, and the one filed as Major was pre-existing.

`composer check` green: 1641 tests / 2727 assertions, 18 property tests, PHPStan clean.

PR #145, opened 2026-08-26
perf(helpers): resolve each link URL once per language

parisek:
## From which project

`sloneek` (redesign branch), while profiling why an uncached page render takes ~3.1 s on the Cloudways box that becomes production.

## Why

`Helpers::formatLink()` calls `url_to_postid()` for every link it formats. WPML filters that call through `SitePress::url_to_postid` → `AbsoluteLinks::_process_generic_text` → `WPML_Get_Page_By_Path::get` → `get_page_by_path()`, i.e. a database lookup per call. A miss costs two, because the slug fallback calls `url_to_postid()` again.

## Measured end to end

Re-measured after the rebase, because the rebase changed the answer. Not extrapolated — the live site was patched and re-profiled, then reverted with the vendor md5 verified identical.

Uncached render of `/cs/`, nine baseline requests in two separate sets and four with the change:

| | before | after |
| --- | ---: | ---: |
| `url_to_postid` calls | 272 | **35** |
| distinct URLs | 29 | 29 |
| SQL queries | 609 | **583** |
| render, median | 2846 ms | **2596 ms** |
| render, mean | 2866 ms | 2590 ms |

**The saving is ~250 ms of a ~2.85 s render, about 9 %.**

### Why this figure is not the one this description used to carry

It said ~37 ms, about 1.2 %, and that was honest for the code as it stood. The rebase onto `main` changed the shape: this branch caches **two** things — the finished link URL and the post id — and `main` had meanwhile extracted the resolution into `urlToPostId()`. Rebuilding the memo around the extracted version put both layers in the path that gets measured, and the second one removes 26 SQL queries the first left through.

So the direction held twice and the size changed twice. The number to trust is the one measured against the code that will actually merge.

Baselines were taken in two sets on either side of the patched run, so a server that drifted mid-measurement would have shown up: both sets returned 272 calls and 609 queries exactly. The first request after a cache flush is a different animal (45 distinct URLs, 1650 queries, ~4.6 s) and is excluded from both sides.

Run-to-run variance on the timing is ±160 ms, so the ~250 ms is above noise but not precise to the millisecond. The call and query counts are deterministic and carry the claim.

For context on where the rest goes: the same request spends 727–756 ms in one `timber/context` call. The database is not the bottleneck. The context build is filed as #147.

## What it means

`formatLink()` keeps its signature and behaviour. Resolution moves into two private helpers with an in-process memo in front.

- **Key is `<blog>|<language>|<url>`.** `url_to_postid()`, `get_permalink()` and `wpml_object_id` all answer for the current blog and the current language. A `switch_to_blog()` or a WPML language switch changes the correct answer for an unchanged URL, and WPML does switch language mid-request while a language switcher renders.
- **A `null` result is cached too**, so an unresolvable URL does not cost two `url_to_postid()` calls at every later call site.
- **In-process, with an explicit boundary.** For a web request that equals request-scoped. Under WP-CLI or a persistent worker it does not, so `StarterBase` flushes the memo on `clean_post_cache` — the hook that fires exactly when a permalink can have moved — and `Helpers::flushTranslatedLinkUrls()` is public for long-running callers that do not boot StarterBase.

One behaviour fix rides along: `get_permalink()` returning `false` is now treated as unresolved. It was previously concatenated onto, so a failed id turned `…/page/?a=1` into `?a=1`.

### Rejected

- **Caching only the post id.** The permalink lookup is the other half of the cost and is scoped the same way, so caching the finished URL removes both under one invalidation rule.
- **An object cache.** It would need invalidating whenever a permalink, translation link or language setting changes. The measured win is already in the in-process memo; a persistent one buys invalidation risk for it.
- **Dropping the negative cache.** Raised in review as a staleness risk. A miss costs two lookups, and within one unit of work an unresolvable URL does not become resolvable; `clean_post_cache` now covers the long-running case that made it worth raising.

## Verification

- `composer check` green: **1751 tests, 2892 assertions**, PHPStan clean.
- **Output equivalence checked, not assumed.** Eight pages across five languages rendered with and without the change. All eight differed — and so did two runs of the *same* code, by exactly the same line counts, because `uniqueId()` emits a fresh id per request. With those normalised, all eight are byte-identical. For a cache, that is the question that matters.
- End-to-end on a real five-language site, patched and reverted (vendor file restored, md5 verified identical).
- Six new tests: repeated URL resolves once; an unresolvable URL is cached as a miss; the same URL resolves separately in two languages; the same URL resolves separately on two blogs; a flush lets a changed permalink through; a `false` permalink leaves the stored URL untouched.
- `HelpersTestCase::setUp()` flushes the memo. A static outlives the test that filled it, so without this one test's answer reaches the next instead of its mocks being called — this bit during development.

## Downstream

No action needed. The change is internal to `formatLink()`; every caller sees the same output.

**One limit worth knowing.** Both memos are flushed on `clean_post_cache`, which covers saving and deleting a post. It does not cover a permalink-structure change or a WPML translation being relinked. A web request ends before either can happen, so this is only reachable from a long-running process — a WP-CLI command that changes permalink structure and then formats links in the same run would be handed the earlier answer.

https://claude.ai/code/session_01RJfaWNpucdqu4trbqiZdB3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,28 +8,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Changed
 
-- `Helpers::formatLink()` resolves each link URL once per language instead of
-  once per call site. The resolution runs `url_to_postid()`, which WPML filters
-  through `SitePress::url_to_postid` and `AbsoluteLinks` into a
+- `Helpers::formatLink()` resolves each link URL once per blog and language
+  instead of once per call site. The resolution runs `url_to_postid()`, which
+  WPML filters through `SitePress::url_to_postid` and `AbsoluteLinks` into a
   `get_page_by_path()` query — and a miss costs a second lookup, because the
-  slug fallback calls `url_to_postid()` again. Profiling a five-language front
-  page found **310 calls for 33 distinct URLs**: the same menu and options-page
-  links formatted once per place they appear. Memoizing removed ~122 ms of a
-  3.1 s render.
+  slug fallback calls `url_to_postid()` again.
 
-  The cache key carries the current language, because `wpml_object_id` and
-  `get_permalink()` both answer per language and WPML switches language
-  mid-request while a language switcher renders. It is request-scoped rather
-  than persistent: a static array cannot go stale, so nothing has to invalidate
-  it when a permalink changes. `Helpers::flushTranslatedLinkUrls()` exists for
-  tests, which are the only caller that outlives the data.
+  Measured end to end on a five-language front page, by patching the site and
+  re-profiling rather than by extrapolating:
 
-  Behaviour is unchanged for every caller. `get_permalink()` returning `false`
+| | before | after |
+  | --- | ---: | ---: |
+  | `url_to_postid` calls | 310 | 47 |
+  | time in the filter chain | 134.1 ms | 96.7 ms |
+Behaviour is unchanged for every caller. `get_permalink()` returning `false`
   is now treated as an unresolved URL instead of being concatenated onto.
 ## [1.42.0] - 2026-08-26
 
-### Added
+  310 calls covered **33 distinct URLs** — the same menu and options-page links
+  formatted once per place they appear. The 47 remaining are those 33 plus a
+  slug-fallback second lookup for the 14 that resolve to nothing.
 
+The saving is **~37 ms of a 3.1 s render**, about 1.2 %. An earlier draft of
+  this entry claimed ~122 ms by scaling the per-call cost; measurement showed
+  the eliminated calls were the cheap ones, because WordPress and WPML already
+  cache parts of the path internally. The expensive work is the distinct
+  resolutions, which remain.
+
+  The cache key carries the blog id and the current language. `url_to_postid()`,
+  `get_permalink()` and `wpml_object_id` all answer for the current blog and the
+  current language, so `switch_to_blog()` or a WPML language switch changes the
+  correct answer for an unchanged URL.
+
+  The cache is in-process. For a web request that equals request-scoped, because
+  the process ends before a permalink can change. Under WP-CLI or a persistent
+  worker it does not, so `StarterBase` flushes the memo on `clean_post_cache`,
+  and `Helpers::flushTranslatedLinkUrls()` is public for long-running callers
+  that do not boot StarterBase.
 - `$breeze_warmup_tail` — keep warming the URLs the cap excluded, a batch at a
   time, in score order, pausing whenever Breeze is draining its own preload
   queue. Batch size is `$breeze_warmup_tail_batch` (default 100 per five-minute
@@ -148,6 +163,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- `Helpers::formatLink()` no longer concatenates a query string or fragment onto
+  `get_permalink()`'s `false` return. An id that failed to resolve produced a URL
+  like `?a=1`; the link is now left as stored.
 - `Helpers::formatLink()` no longer replaces a valid link with an empty string.
   `get_permalink()` answers `false` for a trashed post or a stale WPML
   translation id, and the concatenations after it coerced that to `''` — so a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Changed
+
+- `Helpers::formatLink()` resolves each link URL once per language instead of
+  once per call site. The resolution runs `url_to_postid()`, which WPML filters
+  through `SitePress::url_to_postid` and `AbsoluteLinks` into a
+  `get_page_by_path()` query — and a miss costs a second lookup, because the
+  slug fallback calls `url_to_postid()` again. Profiling a five-language front
+  page found **310 calls for 33 distinct URLs**: the same menu and options-page
+  links formatted once per place they appear. Memoizing removed ~122 ms of a
+  3.1 s render.
+
+  The cache key carries the current language, because `wpml_object_id` and
+  `get_permalink()` both answer per language and WPML switches language
+  mid-request while a language switcher renders. It is request-scoped rather
+  than persistent: a static array cannot go stale, so nothing has to invalidate
+  it when a permalink changes. `Helpers::flushTranslatedLinkUrls()` exists for
+  tests, which are the only caller that outlives the data.
+
+  Behaviour is unchanged for every caller. `get_permalink()` returning `false`
+  is now treated as an unresolved URL instead of being concatenated onto.
 ## [1.42.0] - 2026-08-26
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -250,6 +250,10 @@ The class is `final` with three public static methods: `render()`, `isInserterPr
 
 `BlockRenderer::flushPostBlockCache($post_id)` is the handler `StarterBase` wires to `acf/save_post` at priority 20. When ACF saves a post, the cache group `acf_block_{$post_id}` is flushed — invalidating exactly the cached blocks tied to that post without touching others. The handler guards against non-numeric ids (ACF options-page strings, opaque `block_*` ids) and against environments without `wp_cache_supports('flush_group')`.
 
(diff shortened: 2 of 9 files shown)


parisek:
## Agent review — Codex (via `codex-rescue`), 2026-08-26

Adversarial review of the first commit, asked to challenge the cache-key scope, the negative caching, the `get_permalink()` change, the static lifetime claim and the test quality. Summarised, not pasted. Verdict was "merge with fixes"; both fixes are in `8706bee`.

**Fixed — both confirmed against the code before acting:**

1. **Cache key omitted the multisite blog id.** `url_to_postid()`, `get_permalink()` and `wpml_object_id` answer for the current blog, so a `switch_to_blog()` between two calls would return the other site's permalink for an unchanged URL. Key is now `<blog>|<language>|<url>`.
2. **"The process ends first" was false.** It holds for a web request and not for WP-CLI or a persistent worker, where one process runs many units of work — a command that renames a slug and then formats a link would get the pre-rename permalink. `StarterBase` now flushes the memo on `clean_post_cache`, and `flushTranslatedLinkUrls()` is documented for callers that do not boot StarterBase. This was the assumption the whole design rested on, so it was the finding worth having.

**Rejected:**

- **Drop the negative (`null`) cache**, on the grounds that a miss can mask a later-resolvable URL. Within one unit of work an unresolvable URL does not become resolvable, and a miss costs two `url_to_postid()` calls, so caching it is where the negative case pays. The long-running case that made this worth raising is now covered by `clean_post_cache`.
- **"Behaviour unchanged for every caller" is inaccurate** — correct, and the description said so about `get_permalink()` already. No code change; the phrasing is now scoped to the resolution path.

**Reviewer error worth recording:** the review could not reach the GitHub API and reviewed `main...HEAD` locally instead, so it judged the diff without the PR description. Its complaint that the changelog overstated the scope claim was therefore aimed at text the description had already qualified. The two confirmed findings stand on their own and were verified in the source.

**Not from the review, found by testing:** the claimed saving was wrong. The first description said ~122 ms, scaled from the per-call cost. Patching a real site and re-profiling gave 310 → 47 calls and 134.1 → 96.7 ms, so ~37 ms of a 3.1 s render. The eliminated calls were the cheap ones. Description and changelog now carry the measured figures.

Only one reviewer ran on this. A second, with a different lens, would be worth it before this leaves draft.

parisek:
## Base moved: 1.40.0 landed while this was open

`main` released 1.40.0 mid-review and rewrote the same lines. Both consequences are resolved in main's favour in `c0025c5`.

**The seam moved, and main's is better.** 1.40.0 extracted `Helpers::urlToPostId()` — `url_to_postid()`, the language-prefix fallback behind an `isSiteUrl()` gate, and the `wpml_object_id` translation — as a shared primitive with other callers. The memo now sits in front of *that*, not in front of a private helper invented inside `formatLink()`. Every caller gets it, the diff against main is smaller, and the measurement still applies unchanged because the filter chain profiled is exactly what `urlToPostId()` runs.

**The `get_permalink() === false` fix is no longer mine.** 1.40.0 added it independently, with the same reasoning about a bare `?query`. That is a convergence, not a conflict — both sides wrote the same lines. The description and changelog no longer claim it.

**Test fallout worth stating.** `UrlToPostIdTest` and `Breeze\CuratedUrlsTest` extend `TestCase` directly, not `HelpersTestCase`, so the flush added earlier did not reach them and four of their tests were being answered by an earlier test's mocks. Both now flush in `setUp()` and mock `get_current_blog_id`.

That is the honest cost of memoizing a public static: every test that touches it has to reset it, including tests written later by someone who does not know the memo exists. An injected cache would avoid it; that was rejected as too large a change to a helper this widely called. Stating the cost here rather than letting it be discovered.

## CI is not running on this branch any more

`composer check` is green locally — **1677 tests, 2778 assertions**, PHPStan clean.

But GitHub has not started a `Tests` run for either `8706bee` or `c0025c5`: `actions/runs?head_sha=c0025c5` returns `total_count: 0`, while Actions are enabled and the run for the first commit (`1e7d311`) passed. Only `Lint PR Title` has fired since. I could not find the cause from the API and did not want to force it by reopening the PR or pushing an empty commit.

**So the last two commits are verified locally and not by CI.** Worth a look before this leaves draft.

parisek:
## Agent review #2 — `pr-review-toolkit:code-reviewer`, 2026-08-26

Second reviewer, deliberately given a different lens from Codex: downstream impact, public API surface and maintainability for a shared package, rather than correctness of the cache key. It ran the suite under `--order-by=random`, which the first reviewer and I did not.

**It found a real defect both earlier passes missed. Fixed in `bac6431`.**

**1. A third caller.** `Breadcrumb::build()` resolves menu item URLs through `Helpers::urlToPostId()` (`Breadcrumb.php:447,457`), and `BreadcrumbTestCase` extends `TestCase` directly, so the flush added for the other two never reached it. My commit message said "Both now flush" — true of the two files it named, and it read as a claim of completeness it had not earned. `BreadcrumbTestCase` now flushes and mocks `get_current_blog_id`.

**2. Undocumented public API.** `flushResolvedPostIds()` had no README entry, while the package documents `BlockRenderer::flushPostBlockCache()` in two places. The WP-CLI/worker integrator this method exists for had no way to find it. Now documented beside it, following that precedent.

**Where I disagree with the severity, with numbers.** The review calls #1 blocking on the grounds that the suite breaks under random order. It does — but `main` already fails **31 tests** that way, for unrelated missing mocks (`get_post_type`, `wp_next_scheduled`, `acf_decode_post_id`, `get_field`). The suite was order-dependent before this branch existed. So the bar this PR has to clear is not "make the suite order-clean", it is "do not add to the pile":

| | random-order errors | `get_current_blog_id` failures |
| --- | ---: | ---: |
| `main` | 31 | 0 |
| this branch, before `bac6431` | 31 + new | several |
| this branch, after `bac6431` | 30–31 | **0** |

Repo-wide order-independence is a real problem and a separate one.

**Rejected:** making `BreadcrumbTestCase`, `CuratedUrlsTest` and `UrlToPostIdTest` extend `HelpersTestCase`. The reviewer is right that it is the better shape and would stop a fourth caller repeating this. It rewrites three test bases this branch does not otherwise touch, and the problem it addresses is wider than this memo. Worth its own PR.

**Where the two reviewers converged:** nowhere, and that is worth saying. Codex found two correctness defects in the cache key and the lifetime assumption; this one found a missed call site and a documentation gap. Three real findings across two reviews, zero overlap — which describes the lenses, not the findings. A correctness-only pass would have shipped an undocumented API and a broken third test base; an impact-only pass would have shipped a cache that returns another site's permalink.

## CI

The earlier note said CI had stopped running on this branch. **Cause found: the PR was conflicting.** GitHub cannot build the merge ref for a conflicting PR, so `pull_request` workflows never start — which is also why the other open PR showed only `Lint PR Title`. Resolving the conflict with `main` started them again immediately. Not a repo or billing problem.

parisek:
## Rebased onto main, and one thing it would have undone

`main` moved by #141, #150 and #151 — all three inside the code this PR memoizes — so this needed more than a conflict resolution.

**The trap was `resolveLinkUrl()`.** It carried a copy of the inline resolution as it looked when the branch was cut: raw `url_to_postid()`, the slug fallback, a two-argument `wpml_object_id`. `main` has since extracted exactly those steps into `urlToPostId()` and given them three fixes. Keeping the copy would have silently reverted all of them:

| Fix | What returns without it |
| --- | --- |
| Same-host gate on the slug fallback (#141) | a foreign URL can borrow a local path |
| Translation targets the URL's language (#150) | every translated link resolves to its default-language sibling |
| Language host trusted only where hosts discriminate (#151) | every link resolves to whichever language WPML lists first |

It delegates now, so there is one resolution and the memo sits in front of it.

**The post-id memo needed rebuilding rather than choosing.** This branch caches two things — the finished link URL and the post id — and the rebase kept `main`'s `urlToPostId()` for the second, dropping its cache with it. The wrapper is back around `main`'s body.

## One judgement call worth flagging

The post-id key still carries the current language, and the comment now says why that is conservative rather than necessary. Since the translation targets the URL's own language, the final answer should no longer vary with the ambient one — but the intermediate `url_to_postid()` lookup still does, and that the two cancel out is a reasoning chain, not a measurement. Over-keying costs a repeat lookup; under-keying would hand a language switcher one language's answer for every entry. Dropping the language from that key is a measurable follow-up, not something to do on a hunch.

## Test changes are consequences, not housekeeping

Two suites needed `get_current_blog_id()` defined, because the key reaches for it. `FormatLinkTest::test_unresolved_url_is_cached_too` needed `home_url()`: the slug fallback is now gated on the URL being this site's, so without it the gate cannot answer, the fallback never runs, and the test would assert **one** lookup where a miss really costs **two** — green, and measuring the wrong thing.

`composer check` green: 1751 tests / 2892 assertions, PHPStan clean. All six checks pass and the PR is mergeable.

PR #146, opened 2026-08-26
feat(acf): keep local JSON on disk when a group is deleted in wp-admin

### LINKED ISSUE #144: Keep local JSON on disk when a field group is deleted in wp-admin
## Summary

Deleting an ACF field group in wp-admin also deletes its `acf.json` file from the theme directory. The kit routes those files into the theme itself (`StarterBase::acf_json_save_paths()`), so the kit owns this risk. Keep the database delete. Stop the disk delete.

## What happened

On 2026-08-26 a field group was deleted from wp-admin on a staging site. ACF removed `wp-content/themes/<theme>/templates/page/acf.json` from disk. That file is the only definition of `group_page_header`, which carries nine fields:

```
header_style, header_position, page_hide_header, page_hide_footer,
page_hide_breadcrumbs, page_breadcrumbs_title,
header_primary_button, header_secondary_button, show_exit_popup
```

The visible symptom was one landing page that rendered a header and a footer it must hide. The real scope was every page on the site, because all nine fields fell back to their defaults at once.

Nothing reported an error. The site returned HTTP 200. The postmeta still held the correct values.

## Why nothing complained

`get_field_objects()` resolves each meta key through its `_<key>` reference and skips a reference it cannot resolve (`includes/api/api-template.php`):

```php
$field = acf_get_field( $meta[ "_$key" ] );

// bail early if no field, or if the field's name is different to $key
if ( ! $field || $field['name'] !== $key ) {
    continue;
}
```

A plain `continue`. No notice, no `_doing_it_wrong()`. For ACF this is a normal state, because postmeta outlives field groups that somebody removed on purpose.

`Helpers::formatFields()` builds its array from that result, so the key is absent rather than `false`. Consumer code then reads an absent key:

```php
return ! empty( $page_fields[ $field ] );   // absent -> false
```

That is legal PHP. `WP_DEBUG` stays silent. The value is not lost, it becomes unreadable.

This is the same ambiguity `isFormattedFieldPresent()` already addresses one layer lower. That fix separates `false` from absent. It cannot help when ACF drops the field before `formatFields()` ever sees it.

## Mechanism

`includes/local-json.php` registers two listeners in its constructor:

```php
add_action( 'acf/trash_field_group',  array( $this, 'delete_field_group' ) );
add_action( 'acf/delete_field_group', array( $this, 'delete_field_group' ) );
```

`delete_field_group()` reaches `delete_file()`, which unlinks the file when it is writable.

`acf_delete_internal_post_type()` calls `delete_post( $id )`. A group that lives only in JSON has an id of `0`, so there is no database row to remove. For that case the file delete is the only lasting effect of the whole operation.

## Proposed change

Remove both listeners from `acf/init`, next to the existing `acf_json_save_paths()` wiring in `StarterBase`:

```php
$json = acf_get_instance( 'ACF_Local_JSON' );
remove_action( 'acf/trash_field_group',  array( $json, 'delete_field_group' ) );
remove_action( 'acf/delete_field_group', array( $json, 'delete_field_group' ) );
```

`acf/init` is the correct timing. `acf.php` includes `local-json.php` at line 253, which constructs the instance, and fires `acf/init` at line 489.

The result is the wanted behaviour. A delete in wp-admin removes the database copy. The JSON file survives. ACF registers the group again from that file on the next load, so the group reverts to the version in git instead of disappearing.

## Not verified

The `remove_action` fix is read from the ACF source. It is not tested. ACF may synchronise a JSON-only group into the database before it lists it, and the behaviour would then differ. Test on a throwaway environment before merging.

## Rejected

- **File permissions `444`.** `delete_file()` unlinks only when `is_writable()` passes, so read-only files survive. A deploy rewrites the mode, which is how the restored file in this incident arrived as `664`. Not a control we own.
- **An orphan detector.** A WP-CLI command can compare every `field_*` reference in postmeta against `acf_get_field()` and report the ones nothing defines. It works, and it does not read the deleted file. Measured against production data for this incident, it surfaces all nine lost fields with their row counts, against a noise floor of 84 known legacy keys. It is still the wrong tool here, because it reports after the damage. Prevention is what this issue asks for. Filed separately if we want it.

## Scope

`StarterBase` only. No migration. No change to `Helpers`.

parisek:
Closes #144.

## What changed

`$acf_json_keep_on_delete` (opt-in, default off) removes the six `ACF_Local_JSON` delete listeners on `acf/init`. A delete in the ACF admin still removes the database record. The JSON file stays, ACF registers it again on the next load, and the group reverts to the committed version instead of disappearing.

Six listeners, not the field-group pair alone: `acf_json_save_paths()` routes post-type and taxonomy JSON into the theme as well, so guarding only field groups leaves the same hole one door along.

Default off per the feature-flag rule in AGENTS.md — this changes what a delete in wp-admin does, which is exactly the kind of thing a consumer could be surprised by.

## Verification

The issue filed the fix as read-from-source and untested. That caveat is now retired. Probed against a live ACF 6.8.7 + ACFML install, with the save path forced to a scratch directory:

```
control          written: acf.json   after delete: deleted
guard applied    written: acf.json   after delete: SURVIVED (acf.json)
```

Three further facts the probe pinned:

- All six hooks carry the callback names the implementation assumes, on real ACF 6.8.7.
- ACFML registers its own listeners on the same hooks (`FieldHooks::deleteFieldGroupPackage`, `CptHooks::delete`, `TaxonomyHooks::delete`). They clean up WPML string packages, are unrelated to the file, and survive the guard. Confirmed by re-reading the hook after the removal.
- `group_page_header` on the origin project resolves with `ID` `0` and `local` `json`. A JSON-only group registering with no database row is what makes the revert behaviour work.

`composer check` is green: 1680 unit + 20 property tests, PHPStan level 8 clean, ADR index consistent.

## Rejected

- **File permissions `444`.** `delete_file()` unlinks only when `is_writable()` passes, so a read-only file does survive. A deploy rewrites the mode — that is how the restored file in the originating incident arrived as `664` — so it is not a control this package owns.
- **An orphan detector.** A WP-CLI command can compare every `field_*` reference in postmeta against `acf_get_field()` and report what nothing defines. Measured against the origin project's production data it surfaces all nine lost fields with row counts, against a noise floor of 84 known legacy keys. It works, and it does not read the deleted file. It still reports after the damage, and the issue asks for prevention.
- **Removing ACFML's listeners too.** Same hooks, unrelated job.

## Not covered

The `! function_exists( 'acf_get_instance' )` early return has no test. Brain\Monkey keeps mocked functions defined for the rest of the run, so the branch is unreachable once any earlier test has mocked it — the gotcha AGENTS.md already documents. Covered by inspection.

`$acf_datastore` is still undocumented in README. Found while adding the ACF section; left alone rather than bundled in.

## Rebase

The branch was cut from `feat/breeze-warmup-tail` by mistake, so the PR against `main` originally carried 16 commits and 19 files — it had swallowed the unmerged branch of PR #138. Rebased onto `main`: 3 commits, 5 files, no warmup content. The `CHANGELOG.md` conflict was resolved by hand, and the second review pass caught that the resolution had moved this PR's entry from `### Added` into `### Fixed`. Fixed in 3f486a6.

## CI

No workflow run was created when this PR opened, on the first `synchronize`, or after a close and reopen — a dropped event on GitHub's side rather than anything in the branch. The force-push that published the rebase produced the event that had been missing. All checks now pass.

## Review

Two agent passes, recorded in the comments: Fable on ACF runtime correctness (no findings, independently re-derived the hook behaviour from the ACF 6.8.7 source), Codex on API design and repo doctrine (two findings in pass one, two more after the rebase, all four fixed, none rejected).

https://claude.ai/code/session_01795RYHc999HhovCnadipLz


parisek:
Reopening to trigger the workflow run — no `pull_request` event reached Actions when this PR opened.

parisek:
## Agent review record

Two independent reviewers, 2026-08-26, each given a different lens and no sight of the other's output.

| Reviewer | Lens | Findings |
|---|---|---|
| Claude Fable (`claude-fable-5`) | ACF runtime correctness, WordPress integration | none |
| Codex (`codex-rescue`) | Public API design, downstream impact, repo doctrine | 2, both documentation |

**They converged on nothing.** That describes the lenses, not the findings — the two were deliberately disjoint, and neither strayed. It also means neither finding below has a second opinion behind it; both were checked against the file before acting.

### Fixed (04d88c2)

**Codex, MEDIUM — README over-recommended the flag.** The enable guidance said the flag suits "every project using this package". That is a blanket default dressed as advice, and it contradicts the opt-in framing one paragraph above it as well as the "opinionated defaults live downstream" rule in AGENTS.md. Rewritten around the condition that actually decides it: whether the committed JSON or the database row is the source of truth. The section now also states the case for leaving the flag off, which it did not before — with the flag on, a delete meant to remove a definition for good cannot complete from the admin, because the file registers the group again on the next request.

**Codex, LOW — the recovery claim was wrong, not merely overstated.** The text said a `git checkout` is "the only way back". The reviewer read this as understating the alternatives (backups, other git operations). The real defect is the opposite direction and worse: a group created in the admin and not yet committed has no copy in git at all, so the sentence is confidently wrong in exactly the case where a reader most needs it to be right. Fixed to state both halves. Recording the difference because the next reader is calibrating how far to trust the rest of that report — the finding was real, its reasoning was incomplete.

### Rejected

None. Both findings were valid.

### Noted, no action

Codex could not confirm from a single squashed commit whether the tests were written red-first, and said so rather than assuming. They were: `AcfJsonKeepLocalOnDeleteTest` was run against a missing method first, and the flag-gate cases were added after the wiring existed but before it was believed.

### What the correctness review independently confirmed

Fable did not take the commit message's word for the ACF behaviour and re-derived it from the ACF 6.8.7 source. Worth recording because these are the load-bearing assumptions:

- `local-json.php:51-62` registers exactly the six listeners the implementation targets. None extra, none missed.
- `acf.php:255` includes `local-json.php`, which constructs the instance at file-load time — not gated by `is_enabled()` or the `json` setting, and not itself hooked to anything. `do_action( 'acf/init' )` fires at `acf.php:489`, at the end of the same method. So the listeners are guaranteed to exist when the guard runs, and default priority is sufficient because nothing re-registers them.
- `acf_get_setting( 'json' )` being false does not break the guard: `is_enabled()` gates method bodies, not construction, so the removals are harmless no-ops in that state.
- `acf/delete_*` and `acf/trash_*` fire only from the centralised `ACF_Internal_Post_Type::delete()` / `trash()` methods, used by both the admin UI and the bulk action. There is no separate sync-screen or options-page path that unlinks JSON outside these hooks in 6.8.7.

## CI

No workflow run exists for this branch. `Tests` and `Lint PR Title` both trigger on plain `pull_request` with no `paths` filter and no draft condition, and PR #145 — also a draft, opened the same minute — got both. This branch got neither, on the `opened` event, on a later `synchronize`, or on a `reopened` event after a close/reopen. GitHub Actions reports operational and the workflows are all active; the only check suite on the head commit is the Claude app's, queued. This looks like a dropped event on GitHub's side rather than anything in the branch. Locally `composer check` is green: 1680 unit tests, 20 property tests, PHPStan level 8 clean, ADR index consistent.

parisek:
## Agent review record — second pass

Codex (`codex-rescue`), 2026-08-26, resumed with its own context and shown both its earlier findings and the rebase. Fable was not re-run: the rebase left `src/StarterBase.php` and both test files byte-identical (verified by diffing the pre- and post-rebase hunks), and the only changes since its pass were to `README.md` and `CHANGELOG.md`, which are outside the correctness lens it was given.

### Fixed (3f486a6)

**MEDIUM — the rebase moved this PR's changelog entry out of `### Added` into `### Fixed`.** Real, and mine. The manual conflict resolution anchored on the `## [1.39.0]` heading, which is the *end* of `[Unreleased]`, so the entry landed in whichever subsection happened to be last rather than the one it belongs to. It was correct before the rebase. `release.yml` reads this file, so the flag would have shipped announced as a bug fix. Nothing caught it locally — `composer check` was green throughout, because a changelog entry in the wrong subsection is still a well-formed changelog.

**LOW — the recovery sentence in README was wrong for the common path, not merely absolute.** The reviewer read "has no copy anywhere, and is simply gone" as overstated because backups exist. That is true but is the weaker objection. The sharper one it did not make: ACF hooks `acf/untrash_*` to the save path (`local-json.php:52`), and the admin trashes before it deletes, so a group whose database record was only trashed gets its file written again when the record is restored. Verified in the ACF 6.8.7 source rather than taken on the reviewer's word. Rewritten by case: a committed file comes back from git; a trashed record restores the file on untrash; a JSON-only group has no record to restore, so if it was never committed it has neither. The last case is the one the originating incident hit.

### Confirmed resolved

The first pass's MEDIUM (README recommending the flag for every project) — the recommendation now conditions on which copy is the source of truth.

### Independently verified, not a finding

The off-case claim the revised README adds — that with the flag on, a delete meant to remove a definition for good cannot complete from the admin, because the file re-registers the group on the next request — was checked against the implementation and holds. It was volunteered rather than requested, so it needed checking.

The rebase itself: no entry belonging to `main` was dropped, duplicated or moved, and no content from the `feat/breeze-warmup-tail` branch leaked in. `origin/main..HEAD` is exactly 3 commits and 5 files.

### Rejected

None across both passes. Every finding either was real or, in the LOW case above, pointed at a real defect for an incomplete reason.

PR #152, opened 2026-08-26
fix(media): keep cached derivatives a sibling attachment still uses

parisek:
## From project

`sloneek`, staging. The homepage hero rendered with a black background. The
derivative under `wp-content/cache/image/900x0-center/homepage-hero-desktop.avif`
had been deleted and regenerated by a broken ImageMagick build — but it should
never have been deleted at all: the file, five attachment rows and the rendered
page all still referenced it.

## Why

`cleanup_cached_images()` runs on `delete_attachment` and matches the cache by
basename alone. The cache is keyed by **file**; one file routinely carries
several **attachment rows**:

- WPML writes one row per language over a single file.
- A duplicate upload can be pointed at a path that already exists.

Deleting any one of those rows wiped the derivatives every other row was still
using.

Measured on that site:

```
files shared by more than one attachment : 5 542
attachment rows over them                : 25 981
rows sharing the hero's file             : 5
```

Roughly four fifths of the media library could take a live image down with it.
The failure is silent — no error, nothing in a log, and the damage only shows up
when someone looks at the page.

## What changed

The delete runs only when no other attachment points at the same
`_wp_attached_file`. Comparison is on that meta value because it is what
siblings share; `get_attached_file()` is filtered and absolute, so it is not
comparable across rows.

The guard **fails closed**: where the question cannot be answered, the files are
kept. A stale derivative is overwritten by the next resize; one deleted in error
vanishes from a page that is still serving it.

## Tests

Written first, red for the right reason, then green (`tests/Unit/StarterBase/CleanupCachedImagesTest.php`):

| Test | Before |
| --- | --- |
| deletes derivatives when no other attachment shares the file | passed (behaviour preserved) |
| keeps derivatives when another attachment shares the file | **failed** — files deleted |
| keeps derivatives when the sibling count is unavailable | **failed** — files deleted |

`composer test:all` 1695 + 18 green, `composer phpstan` clean, `composer adr` OK.

## Deliberately not done

- **No `StarterBase` feature flag.** The convention puts behaviour changes behind
  one, default off. Applying it here would leave a data-loss path live for every
  consumer until they opted out of it. Nothing observable is added — a delete
  that destroyed in-use files stops happening. Flagging this is the reviewer's
  call to overturn; I did not want to make it silently.
- **The basename collision is untouched.** Two different files that share a
  basename across upload-year folders (`2022/03/11.png` and `2022/10/11.png`)
  collide in the flat cache namespace, so deleting one still removes the other's
  derivative. 336 basenames on that site map to more than one file. That needs a
  cache-naming change, not a guard, and it deserves its own discussion — the
  fix invalidates every existing cached derivative.
- **`%i` instead of interpolating `$wpdb->postmeta`**, which keeps the query
  literal for PHPStan rather than buying silence with an ignore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `StarterBase::cleanup_cached_images()` deleted resizer derivatives that other
+  attachments were still using. The cache under `wp-content/cache/image` is
+  keyed by file, but one file routinely carries several attachment rows: WPML
+  writes one per language, and a duplicate upload can be pointed at a path that
+  already exists. Deleting any one of those rows wiped the shared derivatives,
+  and the site then regenerated them — or, where the encoder was broken, served
+  a damaged image.
+
+  Measured on a five-language site: 5542 files were shared by 25981 attachment
+  rows, so roughly four fifths of the media library could take a live image
+  down with it. The homepage hero lost its derivatives while five attachment
+  rows and the rendered page still referenced them.
+
+  The delete now runs only when no other attachment points at the same
+  `_wp_attached_file`. It fails closed: where that question cannot be answered
+  the files are kept, because a stale derivative is overwritten by the next
+  resize while one deleted in error vanishes from a page still serving it.
+
+  Not fixed here, and tracked separately: two different files that share a
+  basename across upload-year folders still collide in the flat cache
+  namespace, so deleting one can remove the other's derivative. That needs a
+  cache-naming change, not a guard.
+
 ## [1.41.1] - 2026-08-26
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   `_wp_attached_file`. It fails closed: where that question cannot be answered
   the files are kept, because a stale derivative is overwritten by the next
   resize while one deleted in error vanishes from a page still serving it.
+  "Cannot be answered" covers a missing `$wpdb`, an empty `_wp_attached_file`
+  (a filter supplied the path, so siblings have no key to match on), and a
+  failed query — `get_var()` reports an error by returning null, which casts to
+  the same zero a genuine "no siblings" answer gives, so the null and
+  `last_error` are both checked rather than read as a count.
 
   Not fixed here, and tracked separately: two different files that share a
   basename across upload-year folders still collide in the flat cache
--- a/src/StarterBase.php
+++ b/src/StarterBase.php
@@ -3846,6 +3846,15 @@ class StarterBase extends Site {
 			return;
 		}
 
+		// The cache is keyed by file, not by attachment, and one file routinely
+		// carries several attachment rows: WPML writes one per language, and a
+		// duplicate upload can be pointed at a path that already exists. Those
+		// rows share one set of derivatives, so deleting one of them must leave
+		// the files alone.
+		if ( $this->attached_file_is_shared( $attachment_id ) ) {
+			return;
+		}
+
 		// Extract filename without path
 		$filename = basename( $file_path );
 		$path_info = pathinfo( $filename );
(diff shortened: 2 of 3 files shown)


parisek:
## Agent review — Codex (gpt-5-codex), 2026-08-26

Ran as an independent adversarial pass over the diff. Four findings; two were real
defects in my own guard, two were test-quality. Every one verified against the code
before acting.

### Fixed

**The fail-closed guard was not closed.** `get_var()` reports a failed query by
returning `null`, and `(int) null` is the same `0` a genuine "no siblings" answer
gives. So any transient database error read as *not shared* and deleted the files
the guard exists to protect — the exact failure the PR is about, reachable through
a different door. The docblock already promised the opposite behaviour. Null and
`last_error` are now both checked.

**An empty `_wp_attached_file` returned "not shared".** `get_attached_file()` has
already returned a path by that point, so an empty meta value means a filter
supplied it (offloaded media) and siblings have no key to match on. That is
unanswerable, not negative. Now returns shared.

**The stub validated nothing.** It returned a preselected count whatever the SQL
said, so the suite would have passed against the wrong table, the wrong meta key,
a dropped `post_id != %d`, or misordered args. Added a test pinning the prepared
query and its arguments, plus cases for the two branches above.

**Teardown deleted files it did not create.** `WP_CONTENT_DIR` is a bootstrap
constant, so the cache directory cannot be varied per run — the tests now record
and remove only their own paths. This narrows the blast radius; it does not make
the suite safe to run twice concurrently, and the reviewer was right that only a
per-run directory would.

### Rejected

**`%i` → interpolate `$wpdb->postmeta`.** The finding was that `%i` needs WP 6.2+
and that on older cores the query fails, `get_var()` returns null, and the guard
silently deletes. The premise is right and the *consequence* is what the first fix
above removes: a failed query now keeps the files. What remains on a hypothetical
WP <6.2 is that cleanup stops deleting anything — cache bloat, not data loss, and
visible. The suggested fix would have reintroduced the PHPStan `literal-string`
error that `%i` was chosen to avoid, buying an ignore for no safety gain.

### Worth recording from the confirmations

The reviewer checked the load-bearing premise against local WP 6.8.1 core rather
than asserting it: `wp_delete_attachment()` reads `$file` at post.php:6640 and
fires `delete_attachment` at :6655, before deleting attachment post meta at
:6672–6675. `_wp_attached_file` is present during the callback. If that ordering
were the other way round the whole fix would be a no-op, so this is the single
most useful line in the report.

It also noted the query does not filter `post_type = 'attachment'`, so an orphaned
postmeta row makes retention overly conservative. Left alone deliberately —
erring toward keeping files is the direction this PR argues for.

Tests 1699 + 18 green, PHPStan clean, `composer adr` OK.

parisek:
## Agent re-review — Codex (gpt-5-codex), 2026-08-27

Closing pass of the review recorded [above](#issuecomment-5428174731). Same reviewer, asked to verify its own earlier findings actually landed and then look at the result fresh.

### Verdicts on its earlier findings

- **Fail-closed holes (its blocker)** — ADDRESSED. Verified an empty `_wp_attached_file`, a `null` from `get_var()`, and a populated `last_error` all now return "shared" and keep the files, each with its own test.
- **`%i` needs WP 6.2+ (which I rejected)** — the rejection **holds**, and Codex withdrew the objection: *"I agree with the author's rejection. I no longer disagree."* The premise was right; the dangerous consequence was removed by the fail-closed fix, so what remains on a hypothetical older core is cache retention, not data loss. It suggests documenting WP 6.2 as the effective floor for working cleanup, which is worth doing if older cores are ever supported.
- **Hook-ordering premise** — re-verified against WP core: `delete_attachment` fires at `post.php:6867`, attachment postmeta is deleted at `:6884-6887`. `_wp_attached_file` is present when the callback runs. This is the load-bearing assumption of the whole fix, so it is worth having checked twice.
- **`post_type` filter on the sibling query** — assessed as an acceptable trade, not a defect. An orphaned postmeta row causes over-conservative retention (cache bloat), never data loss, and adding a join buys a failure surface for a case that already fails closed.

### Still open — deferred, not fixed

Three narrow gaps. None is a blocker; all are recorded rather than quietly dropped.

1. **The SQL contract test does not pin the table's position in the query.** It pins the meta key, the `post_id != %d` exclusion and every argument — but because the stub returns the query unchanged rather than substituting placeholders, production could change `FROM %i` to `FROM wrong_table` and the test would still pass.
2. **`seedFile()` owns paths, not files.** Teardown was narrowed to remove only the test's own paths, but ownership is path-based rather than creation-based, so two concurrent runs still share fixture paths. `WP_CONTENT_DIR` is a bootstrap constant, so a genuinely per-run directory is not available; making `seedFile()` fail when its target already exists would close most of it.
3. **Two adjacent docblocks above one test method.** PHP associates only the second with the method, so a description meant for a neighbouring test sits in the wrong place. My error, introduced when the fail-closed tests were added. Documentation only.

Fixing these means new commits on this branch, and #153 is stacked on it — so the rebase belongs to whoever picks that up, and is why they are recorded here rather than pushed unilaterally.

CI 6/6 green.

parisek:
The three findings recorded above as *Still open — deferred, not fixed* are now closed in #163 — tests and comments only, no production code. Both behavioural fixes are mutation-verified there.

PR #153, opened 2026-08-26
feat(resizer): scope the cache key by the source's upload path

parisek:
> **Stacked on #152.** Base is `fix/cleanup-cached-images-shared-file`, not `main`. Merge #152 first; GitHub retargets this one automatically. Do not delete the parent branch until both have landed (AGENTS.md § Stacked PRs).

## From project

`sloneek`. A homepage hero rendered with a black background. The cause turned out to be two independent defects, and this PR is the second: `Resizer` names a derivative after its source's basename alone, so two uploads that share a name share one cached file.

## The measurement

On that site's staging database:

```
names mapping to more than one source file : 496
```

The cache namespace is flatter than the namespace it caches, **twice over** — the derivative is named for the source *without its extension*, so `11.png` and `11.jpg` collide as surely as two `11.png` in different months do:

```
900x0-center/11.avif   <-  2022/03/11.png
                       <-  2022/04/11.png
                       <-  2022/08/11.png
                       <-  2022/10/11.png
```

Whichever renders first writes the file; the rest read it and get a picture of something else. Nothing errors, and the result is a plausible image, so it stays invisible until someone recognises the wrong photograph on a page.

## What changed

The source's own upload directory now sits between the size directory and the filename:

```
<W>x<H>-<style>[-q<N>]/<source-upload-dir>/<name>.<fmt>
```

Behind `StarterBase::$resizer_source_path_in_cache_key`, **default `false`**, mirroring the `$resizer_quality_in_cache_key` flag it sits beside. It is derived from the source URL — not the resolved file path — because `resize()` needs the value before it knows whether the file exists locally: a missing source is handed to DevMediaProxy, which addresses the same cache path on another host.

The segment is written into `$variant['cache_key']`, not into the target path. `cache_key` is already the one value both the local writer and `DevMediaProxy::build_remote_resizer_variant()` derive from, and the comment at `DevMediaProxy.php:456-460` names that as deliberate. Appending in `processVariant()` would have left the remote path in the old shape.

**`wp timber-kit migrate-image-cache`** moves an existing cache into the new shape rather than re-encoding it. Dry-run by default; `--apply` writes. It moves rather than copies (no transient doubling), re-checks the target immediately before every `rename()`, never deletes, never overwrites, and is idempotent so an interrupted run is resumed by running it again.

Derivatives whose name maps to more than one source **cannot be placed** — recovering that association is exactly what the flat layout destroyed. They are reported and left alone, and re-encoded on first view.

**With the flag on, `cleanup_cached_images()` addresses derivatives by computed path** instead of matching basenames across the tree. That closes the delete half of the same bug. The sibling-attachment guard from #152 stays in force — it answers a different question and is independent of the layout.

**A site with year/month folders switched off gets byte-identical paths** and needs no migration at all. The suite pins that.

## Review

Every task was reviewed independently before it landed. Three Critical defects were caught this way and are fixed in the branch, all of the same shape — **a check performed at one moment and relied on at another, where the gap is silent**:

1. `apply()` re-used `plan()`'s conflict check, and POSIX `rename()` silently replaces its destination. A target written between plan and apply was destroyed with no error and no entry in `failed`.
2. The delete path took the filename **raw** from `_wp_attached_file` and used it as a `glob()` pattern, while the writer sanitizes it. Deleting `a[1].png` removed unrelated `a1.png`'s derivative, because `[1]` is a glob character class. Names carrying stripped characters silently no-opped and leaked instead.
3. `buildNameToDirs()` dropped a guard-rejected attachment from the map entirely, which could make a genuine collision read as unambiguous — and move one attachment's derivative into another's directory. The reopened swapped-image bug, inside the tool operators run to make the flag-flip safe.

Each has a regression test that fails against the code as it was.

Three components now implement "is this path component safe to use as a directory" — the writer, the migrator and the deleter. They are deliberately **not** coupled; the final review compared all three against each other, since any input they disagree on is a bug.

## Deliberately not done

- **`Helpers::resizeImage()`** — the legacy path predating `Resizer` — keeps the flat layout. Not migrated, not flagged.
- **Symlinks inside the cache tree** are not handled. The theme owns and regenerates this directory, so it is outside the threat model. Stated rather than silently skipped.
- **The flag defaults off, so the bug stays live until a project opts in.** That is the cost of the change invalidating every cached derivative — it has to be the project's decision, not a consequence of `composer update`. This differs from #152, where I argued against a flag because nothing was at stake.

Decision, rejected alternatives and consequences: `docs/adr/0007-resizer-source-path-cache-key.md`, which lands in this PR.


parisek:
## Agent review — Claude (opus) and Codex (gpt-5-codex), 2026-08-27

Two independent reviewers, different models, neither able to see the other's output. Recorded here because the rejected and deferred findings are the ones the diff cannot explain.

**They converged on nothing.** That is worth stating plainly rather than hiding: it describes the lenses, not the findings. Claude was asked to verify the fix works; Codex was asked to find where it still does not. Both did their job and returned disjoint sets.

### What Codex found that seven Claude review rounds did not

The branch originally added only the source's **directory** to the cache key. Codex found that closes half the problem: the derivative is named for the source *without its extension*, so `2026/08/hero.jpg` and `2026/08/hero.png` still wrote one file, and deleting one removed the derivative the other served.

Measured on production before acting:

```
colliding names total                  : 496
of those, colliding inside one directory : 152
```

So the first version fixed 344 of 496 while the ADR, CHANGELOG and README all claimed the job was done — and the ADR's own opening example (`11.png` vs `11.jpg`) was in the broken half.

The key now carries the source's whole sanitized identity, directory and full filename: `900x0-center/2026/08/hero.webp.avif`. ADR-0007 was amended rather than superseded, since it had not merged.

### Also found and fixed

- **Filename came from the encoded URL while everything else came from the decoded database value.** `sourcePathSegment()` already stripped the query string and decoded each directory component; the filename path did neither, so writer and deleter could compute different names. Measured: **230 `_wp_attached_file` values contain a space** — not theoretical.
- **Root uploads broke the migration's idempotence claim.** A migrated root-upload derivative stays directly inside the size directory, so a second run rescanned it and reported it `orphaned`. Production has zero root uploads; fixed because the library serves other sites and the spec claims idempotence.
- **`apply()` retained a TOCTOU window** between `is_file()` and `rename()`. Now `link()` + `unlink()` — atomic create-if-absent, so a target appearing mid-run fails the move instead of being overwritten.
- **The CLI reported success when moves had failed.** `WP_CLI::success()` ran before the warning loop, so automation got exit code 0 and no detail.
- **The ADR contained six statements the code did not implement**, including a claim that sites without year/month folders need no migration. Adding the extension made that false and I missed it while amending. Corrected.

### Rejected

**Codex rated "sanitization can collapse two distinct source names" as High** — `hero[1].jpg` and `hero1.jpg` sanitize to the same value, so the key is the *sanitized* identity, not the source identity. Its suggested fix was to append a hash of the canonical path.

Rejected, and Codex was asked to challenge the rejection and agreed with it. Measured zero occurrences of `[` or `]` in `_wp_attached_file`, and WordPress sanitizes filenames at upload, so the second sanitisation is normally a no-op. A hash would permanently damage path readability and complicate migration for every site, to guard a case the data does not contain. The limit is now stated in the ADR's Consequences instead — an accepted, documented constraint rather than an accidental gap. Worth revisiting only if unsanitized `_wp_attached_file` values ever surface in a supported install.

### Deferred

- `strtok($src, '?')` does not strip a `#` fragment, where a sibling helper uses `'?#'`. Fails closed; image URLs do not carry fragments.
- `guardSourceDir()` omits the `rawurldecode()` step its sibling performs. Different input domain (`_wp_attached_file` is never percent-encoded); same outcome.
- Symlinks inside the cache tree are not handled — `GLOB_ONLYDIR` descends into a symlinked directory. Outside the stated threat model: the theme owns and regenerates this directory. Stated rather than silently skipped.
- No test covers `link()` succeeding and `unlink()` then failing. The path is handled (recorded as failed, both copies left on disk) but unproven.

### A reviewer error worth recording

Codex's first pass called the `%i` placeholder a silent-delete path on WordPress < 6.2. The premise was right and the consequence was not: the accompanying fail-closed fix means a failed query now keeps the files, so what remains is cache retention, not data loss. On re-examination Codex withdrew the objection. Recorded because the next reader is calibrating how far to trust the rest of that report.

Tests 1750 + 18 green, PHPStan clean, ADR index in sync, CI 6/6.

parisek:
## Rollout notes — measured scope, and what to do when you come back to this

Written for whoever picks this up later, including me. Everything below is measured on `sloneek` production data, not estimated.

### Merging this changes nothing

`StarterBase::$resizer_source_path_in_cache_key` defaults to `false`, and the filter behind it defaults to `false` too. No project sets it. Until one does, the derivative name still drops the source extension, no directory segment is added, and `cleanup_cached_images()` keeps using the original basename scan — a verbatim extraction, not a rewrite.

So this can sit merged and dormant. Nothing reaches a site until someone opts in **and** bumps the package.

Contrast with #152, already merged: that one is **not** behind a flag, deliberately, because gating a data-loss fix would leave the loss live by default. It applies on the next version bump.

### How much is actually broken today

The bug this PR fixes is not hypothetical — it is live right now. But the blast radius is smaller than the raw collision count suggests, and the difference matters for scheduling.

| Measure | Value |
| --- | --- |
| Colliding names in the database | 497 |
| Of those, colliding **inside one directory** | 152 |
| Unique names present in `cache/image` | 1376 |
| **Colliding names that actually have a derivative** | **90** |

**90 is the ceiling on visible damage.** A collision in the database harms nothing by itself; a wrong image can only be served where the shared derivative physically exists. Ninety cache files are currently standing in for two or more different source images each.

### What those 90 are

Almost all of them are auto-generated names — pasted screenshots and import artefacts:

```
10   ->  2022/03/10.png | 2022/08/10.png | 2023/06/10.png | 2024/03/10.webp
11   ->  2022/03/11.png | 2022/04/11.png | 2022/08/11.png | 2022/10/11.png
1634240049156  ->  2023/01/1634240049156.jpg | 2023/03/1634240049156.jpg
```

No `hero`, no `logo`, nothing a human named. Of six colliding sources sampled, **one** appears in published `post_content`, and once.

A looser check over ACF placements (attachment IDs in `postmeta`) returned 500 of 5286 attachment rows — but read both numbers with suspicion: 5286 is inflated by WPML writing one row per language (≈1000 distinct files), and 500 is an upper bound because the test matches any numeric meta value equal to the ID. Neither number raises the ceiling of 90; they only indicate how the affected images are placed.

**Conclusion: real, not acute.** The risk concentrates in inline article images, not in page furniture. That is also why nobody has reported it in three years — one screenshot looks like another, and nothing here gets compared side by side. It is the dangerous category: a failure that renders as a plausible success.

### When you turn it on

Prerequisite: **the Cloudways AVIF encoder must be fixed first.** Its ImageMagick 6.9.11 (2021) drops the alpha channel, so anything regenerated there comes out with a black background — see `docs/migration/cloudways-imagemagick/README.md` in the `sloneek` repo. Migrating before that trades a silent wrong-image bug for a loud black-background one.

Then, in order:

1. Set `$resizer_source_path_in_cache_key = true` in the project's `Base`.
2. Run `wp timber-kit migrate-image-cache` **without** `--apply` and read the numbers. Dry-run is the default for this reason.
3. Expect a non-trivial `ambiguous` count. Those are the derivatives whose source cannot be recovered — precisely what the flat layout destroyed. They are left alone and re-encoded on first view, which is correct and requires a working encoder.
4. Only then `--apply`.

### Unproven, and worth knowing before you rely on it

**The migration command has never run against real data.** It has unit tests over synthetic directory trees and two independent reviews, and three Critical defects were caught and fixed in it — but zero production mileage. The dry-run in step 2 is its first real exercise; treat the first run as an experiment, not a formality.

### Still open on this branch

Three Minor findings from the Codex review of #152 were recorded there and not fixed: the SQL contract test does not pin the table's position in the query, `seedFile()` owns paths rather than files, and two adjacent docblocks sit above one test method. None blocks anything; all are on [#152](https://github.com/parisek/timber-kit/pull/152).

parisek:
## Rebased onto `main` (1.45.0) — ready for testing

`main` moved 11 commits and two releases while this branch sat. Rebased onto `835623e`; CI green on `ac79a88`.

### The ADR changed number: 0007 → 0008

`main` took 0007 for *Prove cache purity from inputs* while this was open, so both files claimed it and the index listed two entries under one number. The decision is unchanged — only the number moves. The spec now lives at `docs/adr/0008-resizer-source-path-cache-key.md`, and every reference in `src/`, `tests/`, the changelog and the plan moved with it.

**Anything citing this work by number needs updating**: it is `ADR 0008` from here on.

### Conflicts, and what each resolution decided

Four, none of them a behavioural change:

| File | Conflict | Resolved |
|---|---|---|
| `docs/adr/README.md` | Both branches added an 0007 index line | Renumbered ours to 0008 |
| `src/Resizer.php` | Our new property sits above a docblock `main` rewrote (#154, backend-format memo) | Kept both — our property, `main`'s new docblock |
| `CHANGELOG.md` | `main` renamed the Unreleased `### Fixed` heading to `### Documentation` | Our `### Added` first, then `main`'s `### Documentation` |
| `CHANGELOG.md` | Our commit amended a #152 paragraph that no longer exists — see below | Dropped the orphaned paragraph; kept our two `### Fixed` entries |

The `Resizer.php` one is worth a second look on review. It is textual adjacency, not logic: #154 added a process-wide memo for the image-backend format probe, which does not touch the cache key. But it is the one place where two branches edited the same lines of the file this PR is about.

### Found while rebasing: #152's changelog entry is missing from `main`

#152 merged with a 30-line `### Fixed` entry in Unreleased (`920890f`). It is **not in 1.44.0, not in 1.45.0, and not in Unreleased** on `main` today — it did not survive release preparation. So a shipped fix has no changelog record.

Not fixed here, because it belongs to `main` and not to this branch. Recorded so it is not lost.

### The precondition on testing is now met

The rollout notes above said: fix the AVIF encoder before enabling the flag. Measured on the production host today:

- ImageMagick **7.1.2-31**, not the 6.9.11 the incident runbook documents.
- A synthetic RGBA PNG encoded to AVIF keeps its alpha — `srgba, alpha=Blend` — both through the CLI and through PHP-Imagick, which is the path the theme actually takes.

So the blocker is cleared. The rollout order in the notes above stands otherwise: set the flag, run `wp timber-kit migrate-image-cache` **without** `--apply`, read the `ambiguous` count, and only then apply.

### Still true, and still the main risk

The migration command has never run against real data. 1937 tests, PHPStan clean, two independent reviewers and five fixed Critical defects — but zero live kilometres. The dry-run is its first real test, not a formality.

parisek:
## Rebased again — onto `main` with #161, #162 and #163

Three PRs landed while this sat in draft, and two of them touch files this branch owns. Rebased onto `dcd2b5c`; CI 6/6 green on `d9befdd`, 1958 tests, PHPStan clean.

### What each one cost

**#163** (closing the three deferred findings from #152) edits `CleanupCachedImagesTest.php`, which this branch also extends. Two conflicts, both the same shape: both sides inserting at the same point, resolved by keeping both. #163's relocated docblock now sits above the missing-database test it describes, and this branch's four flag-on tests sit above it.

Worth stating plainly: **that conflict is a consequence of #163, which I opened.** The three findings were deferred precisely because this branch was stacked on #152 — the price of finally fixing them was this rebase.

**#162** (the resizer output-format health check) adds a `### Added` entry where this branch adds one. Kept both.

One conflict there deserves naming, because resolving it the obvious way would have silently reverted work. The `### Added` paragraph on this branch was **rewritten mid-branch** — the original described the fix as directory-only, the amendment corrected it to the source's whole identity, extension included. At the conflict, `HEAD` carried #162's entry *followed by the superseded paragraph*. Taking "both sides" verbatim would have restored the wrong description and left two entries for one flag. Resolved by dropping the superseded copy; the changelog now carries exactly one, the amended one.

**#161** (restoring #152's lost changelog entry) touches only the released 1.43.0 section. No conflict.

### Re-verified after the rebase, not assumed

#163's two behavioural fixes are mutation-tested, so a rebase that quietly dropped either would leave green tests behind. Both re-checked on this branch:

- Writing the table as a literal instead of `%i` fails the contract test.
- A pre-planted foreign fixture fails `seedFile()`'s ownership guard **and survives teardown**.

### One thing for review, not fixed here

The four flag-on tests added by this branch create their directories with a bare `mkdir()` and push the path onto `$created` by hand — path-based ownership, the exact pattern #163 replaced in `seedFile()`. The class already has `makeDir()`, which records only what did not exist. Switching them is a one-line change per test, but it is a change to this PR's own content rather than a conflict resolution, so it belongs to your review rather than to a rebase.

parisek:
## End-to-end test on a real site (sloneek, local DDEV)

Ran against the real corpus: 15 GB of uploads, 228 610 image files, five WPML languages. The flag was on, the cache emptied first, and 260 pages spidered from the site itself across cs/sk/pl/it.

Two of the four defects below are in my own test harness, not the code. They are recorded because each one first looked like a failure of this PR, and the next person will hit them too.

### Write path

653 derivatives generated from an empty cache. **All 653 landed in the scoped layout; zero flat.** A reload regenerated nothing, so the writer and the reader agree on the key.

### Correctness oracle — the assertion that matters

Every scoped derivative was compared against the source its own path names. The first run reported 57 mismatches. Both causes were the oracle's:

- **36 were crops.** A `1200x640-center` derivative is a crop of its source, so comparing it to the whole source differs legitimately. Fixed by mirroring the resizer's cover-resize plus gravity crop: 57 → 36.
- **The remaining 36 were alpha.** All sat at ~0.457 with suspicious uniformity across unrelated images. Flattening alpha consistently on both sides dropped one measured case from **0.457 to 0.0055** — the same picture.

**Net: zero derivatives depicting the wrong source.**

### Delete path — the half the earlier round never touched

Verified with synthetic attachments, and the first two attempts looked like a bug before the cause turned out to be correct behaviour:

| step | result |
|---|---|
| delete 4 of 5 WPML rows | derivative **kept** |
| delete the last row | derivative **deleted** |
| same-named derivative in a sibling directory | **survived** |

The last row is the point: under the flat layout that neighbour would have gone too.

Worth stating for any WPML site: **every attachment carries ~5 rows**, so `cleanup_cached_images()` only ever deletes on the last language copy. That is #152 working, but it means derivative cleanup is rare in practice here.

### The site filters `sanitize_file_name()`

A plugin on this site rewrites `usp_1.webp` → `usp-1.webp` and `Ebook.svg` → `ebook.svg`. Both the writer and `cached_derivative_paths_by_source_path()` call the same function, so both see the filter and agree — verified end to end with an `E2E_Test.webp` attachment that wrote `e2e-test.webp.avif` and was deleted correctly.

Two consequences worth recording:

1. The cache key is post-sanitisation, so `usp_1.webp` and `usp-1.webp` still collide, as does `Ebook.svg` with `ebook.svg`. ADR 0008 already states this limit; this is it observed in production data, and that exact `Ebook.svg`/`ebook.svg` pair is recorded in the consuming project's own `uploads-transfer.md` as two genuinely different images.
2. Any project whose filter changes while a cache exists silently orphans every derivative written under the old rule.

### Migration

Idempotent: against an all-scoped cache the planner scans 0.

An earlier `--apply` on the same site moved 1634 of 2006 with 0 failures and reported **42 conflicts that the dry run had not predicted** — a page load between the two runs had created those targets. Nothing was overwritten, which is `link()`+`unlink()` doing the job it was written for.

### Finding: non-decodable candidates inflate `ambiguous`

The resizer never writes a derivative for an SVG or PDF source — confirmed, the 653-file cache contains only webp/png/jpg sources. But `buildNameToSourcePaths()` counts SVG and PDF attachments as candidates, so a name whose only *decodable* source is unique is still reported ambiguous.

Measured on this site: **30 of 329 ambiguous entries have exactly one decodable candidate.** Skipping sources the resizer cannot decode would migrate those instead of re-encoding them, and remove 30 false ambiguities from a report a human has to read.

```
0x114-center/corehr-highperformer-highperformer.avif
   candidates: …highperformer.svg, …highperformer.webp
   only decodable: …highperformer.webp
```

Not fixed here — it is a change to this PR's own logic, and it belongs to your review.

### Still not covered

260 pages of a much larger site; `merge_resizer` and mobile variants unexercised; `--apply` never run against live traffic; and this PR still has no independent reviewer.

parisek:
## Two independent agent reviews — Codex (gpt-5-codex) and Fable, 2026-08-31

Different lenses on purpose: Codex on correctness and data loss, Fable on contracts, tests and downstream impact. Neither saw the other's output. Nothing is fixed yet — this records what they found and what I verified against the code and the deployment.

### Where they converged

**1. The migrator's map key can disagree with the legacy writer's key.** Codex called it Major, Fable Minor.

Measured on the deployment rather than argued: for `2023/06/Návrh bez názvu-16.png` the legacy writer key and the migrator key are **both** `navrh-bez-nazvu-16`. They agree, because `wp_get_attachment_url()` returns an unencoded URL here, so `basename()` sees the decoded name on both sides. The 2006-file migration reported exactly **1 orphan**; if this reproduced, every space-bearing name would have orphaned.

So it is real but conditional on something filtering the attachment URL into percent-encoded form. **Fable's severity is the better calibrated one** — the consequence is a re-encode, not a wrong file. Codex reached Major partly by assuming encoded URLs are the normal case.

**2. `sanitize_file_name()` is neither stable nor injective, and the code leans on it for a destructive lookup.** Codex Major #2, Fable verdict (B). This is the strongest signal in the review: same root cause, two lenses, and it matches what the deployment already shows.

**3. `SourcePathCacheKeyTest` does not test what it says.** Both, independently, Minor. Its helper re-implements the append instead of calling `resize()`, so a break in the real wiring passes.

### Fable found the sharpest thing, and it is worse than reported

`buildNameToSourcePaths()` dedupes candidates by the **sanitized** source path (`src/Cli/MigrateImageCacheCommand.php:143`). Two genuinely different uploads whose names differ only in case, spaces or underscores collapse into **one** candidate. The planner then sees an unambiguous name and moves the flat derivative into it — confidently, into a path two different images share.

That is not "the collision is hidden". It converts a genuine ambiguity into a silent wrong placement, in the one bucket the design promises never to guess in.

Quantified on the deployment: **8 collapsing pairs**, every one a distinct pair of files.

```
2023/06/group-1469.png          <= Group-1469.png        | Group 1469.png
2022/01/rychlost-nasazeni.webp  <= rychlost-nasazeni.webp | rychlost_nasazeni.webp
2025/09/team-pict-1.png         <= team-pict-1.png        | team-pict_1.png
2025/12/hr-initiatives-2.jpg    <= hr-initiatives_2.jpg   | hr-initiatives-2.jpg
```

The site filters `sanitize_file_name()` to lowercase and map `_` to `-`, which is what makes these reachable. ADR 0008 dismisses this class as measuring zero occurrences; on this deployment it measures eight, and the ADR does not mention that the function is filterable at all.

### Fable corrected a finding of mine, and the correction matters

I proposed excluding non-decodable sources (SVG, PDF) from the migrator's map, measured at 30 of 329 ambiguities. Fable agrees with the substance but says **implement it against the static `DESIRED_INPUT_FORMATS` set, never against runtime `canDecode()`** — a derivative may have been written while Imagick decoded TIFF or HEIC and the site has since moved to GD, or narrowed `timber_kit_resizer_allowed_types`. Filtering that source out would make the surviving candidate look unique and move **another attachment's derivative** into it.

My suggestion as written would have introduced the exact wrong-placement class this PR exists to remove. Recorded because the fix is only safe in one of its two obvious forms.

### Fable-only, and I confirmed each

- **Major — the property docblock contradicts the ADR amendment.** `src/StarterBase.php:901` still says "A site with year/month folders switched off sees no change at all." False since the extension was folded into the key: a lone root `hero.png` moves from `hero.avif` to `hero.png.avif`. A consumer reading the docblock — the likeliest surface — skips the migration and orphans the whole cache. Same stale claim at `:1400` and in the README hook table; the CHANGELOG is right.
- **Major — no test for flag-on plus a missing source file.** The segment is appended before the `file_exists()` branch so DevMediaProxy addresses the same key; that is a stated design point of the ADR and every flag-on test patches `file_exists` to true. Moving the append below the branch passes the whole suite.
- **Major — `orphaned` and `conflict` are counts with no paths.** I hit this myself: a run reported `conflicts: 42` and `orphaned: 1` and there was no way to learn which files. `--verbose` lists only `ambiguous`. Both lists exist in the plan and are simply not printed.
- Minor — mixed-flag environments are unwarned: a dev site with the flag on, proxying a production origin with it off, requests paths the origin never serves.

### Codex-only

- Minor test-naming gap (converged item 3 above).
- Codex could not run PHPUnit or PHPStan — its sandbox blocked temp files. I ran both on this branch: **1958 tests pass, PHPStan clean.** Its findings are code reading, not test evidence.

### My read

No Critical. The one I would not merge without is the **silent dedupe**, because it is the only finding where the failure is a wrong image placed confidently rather than a re-encode or a stale doc — and it is reachable on a real site today, eight times.

parisek:
## Reworked: the key is the stored name, not a re-sanitized one

This changes the decision ADR 0008 records, so the ADR is amended in the same commit. **Flag off is untouched.** CI green on `06e407f`, 1960 tests, PHPStan clean.

### The question that turned it around

Why sanitize at generation at all? A name is sanitized **at upload** — the stored `_wp_attached_file` value is already what WordPress decided to call the file. Re-deriving a spelling from it, with today's filter stack, throws away information the database had.

Neither the original ADR nor seven review rounds asked that. Both agent reviewers found the *consequences* — Codex the instability, Fable the non-injectivity — and neither questioned the premise.

### What was actually wrong

`sanitize_file_name()` is **filterable**, which the ADR never mentioned. Under a real filter it is:

- **Not stable.** A derivative written before the filter existed is addressed afterwards under a different spelling, so every later lookup — the delete included — points away from what is on disk.
- **Not injective.** Eight collapsing pairs measured on the deployment; **six are visibly different images**, two are duplicate uploads.

The second was worse than an unfixed collision. The migrator deduped candidates by the *sanitized* path, so a pair became one candidate: the name looked unambiguous and the planner moved a derivative into a path two uploads claim — a confident wrong placement in the one bucket that promises not to guess.

### What changed

The key is the stored name **verbatim** (decoded, since the writer reads a URL and the deleter reads the database). The guard **refuses instead of repairing**: a component is used as stored unless it could leave its directory — empty, `.`, `..`, a separator or a NUL. Same rule for both halves; previously only the directory half had it, which is why the asymmetry went unnoticed.

This was only possible because **the delete stopped building a glob pattern**. `glob()` cannot quote a metacharacter — that was the *actual* reason the name was rewritten. Reading the directory and comparing strings removes the hazard and the argument for the rewrite together. `img[0-9].png` is now matched literally, which the rewritten test asserts.

### Two costs, taken knowingly

The derivative URL now percent-encodes each path segment, because a stored name may legitimately carry a space, `#` or `%`. And on a **case-insensitive filesystem** `Ebook.svg` and `ebook.svg` remain one file — the filesystem, not the key, and nothing in this package can change it.

The migrator keeps **both** spellings on purpose: the legacy flat names it reads were written sanitized, so it reproduces that spelling to find them while recording the new identity verbatim.

### Verified on the deployment, not only in tests

- The pair that used to collapse now writes two paths: `Group-1469.png.avif` and `Group%201469.png.avif`.
- Delete round-trip with a name carrying both a space and an underscore — the two characters the site's filter rewrites: written as `E2E Test_Raw.webp.avif`, URL single-encoded, and removed once the last of five WPML rows went.
- The new map-builder regression test is mutation-verified: restoring the sanitize call fails it.

### Also fixed from the reviews

Fable's Major on the stale docblock. `$resizer_source_path_in_cache_key` claimed a site without year/month folders "sees no change at all" — false since the extension joined the key, and the likeliest surface for a consumer deciding whether to run the migration. Corrected in the docblock, the README and the CHANGELOG.

### Correction to my own earlier comment

I wrote that the eight pairs were "each a genuinely different pair of files". I had compared their stored paths, not their pixels. Measured: **six different, two duplicates.** The ADR and CHANGELOG now say six.

### Still open from the reviews

Not addressed here, all previously reported: the non-decodable-source exclusion (and Fable's constraint that it must key off the static format list, never runtime `canDecode()`); `orphaned`/`conflict` printing counts without paths; no test for flag-on plus a missing source; the mixed-flag proxy caveat.

parisek:
## The four remaining review findings are closed

CI green on `851b285`, 1964 tests, PHPStan clean. Each one is measured on the deployment, not just unit-tested.

### 1. A source the resizer cannot decode is no longer a candidate

An SVG or PDF sharing a stem with a real image made that image's derivative look contested, so it was left unmigrated.

Measured before and after on the same cache: **ambiguous 329 → 299.** Exactly the 30 predicted.

Fable's constraint is the load-bearing part and it is now a test in its own right. The exclusion reads `Resizer::inputFormatMimes()` — the **static** superset — never `canDecode()`, which narrows by the live backend and by `timber_kit_resizer_allowed_types`. A site that resized TIFFs under Imagick and has since moved to GD still has those derivatives on disk; dropping the source because *today's* backend cannot read it would leave the surviving candidate looking unique and move another attachment's derivative into it. Both directions carry a test: the SVG is excluded, the TIFF is kept.

### 2. Every bucket that leaves a file behind now names its files

`conflict` carries the target it collides with rather than only the source, and `--verbose` lists conflicts, orphans and the planned moves.

I hit this myself on a real migration — `conflicts: 42` with no way to learn which. Now:

```
conflicts (target already exists -- nothing overwritten):
  0x392-center/speed.avif -> …/cache/image/0x392-center/2024/11/speed.webp.avif

orphaned (no source in the database answers to this name):
  900x0-center/homepage-hero-desktop-5.avif
```

### 3. Flag-on with a missing source has a test

The segment is appended before the `file_exists()` branch so DevMediaProxy addresses the same key from another host. Every other flag-on test stubs `file_exists()` to true, which is exactly why the gap survived.

Mutation-verified: moving the append below the branch fails the new test and nothing else in the suite.

### 4. Mixed-flag hosts are documented

README and CHANGELOG now say the flag must match on every host sharing a cache. A local site proxying a production origin 404s every derivative while the two disagree.

### End-to-end re-run after all of it

Cache emptied, site crawled: 95 derivatives, **zero flat**, and a reload regenerates nothing (11.0s cold → 0.48s warm). `--apply` moved 40 with 0 failures. The migration's own targets keep their spelling — `CoreHR_HighPerformer_Europe_HighPerformer-1.webp.avif`, capitals and underscores intact.

### One consequence worth stating

This rework changes the flag-on layout itself, so a cache built by an **earlier commit of this branch** is addressed under the old sanitized spelling and orphans. That matters to nobody today — the flag has never been released — but anyone testing this branch should empty `cache/image` rather than migrate across the change.

parisek:
## Second review round — on the rework itself

Both reviewers were re-run against `d9befdd..HEAD`, because the rework changed the very mechanisms their first pass had verified. That was the right call: **they found real bugs in it.** CI green on `1044fe0`, 1968 tests, PHPStan clean.

### Codex — Major: the delete over-matched

`strpos( $entry, $name . '.' ) === 0` also matches `hero.png.old.avif`, which belongs to the unrelated source `hero.png.old`. Deleting `hero.png` took another attachment's derivative with it.

**This is the same over-match the old `glob()` had, in a different spelling.** I replaced a pattern language with a prefix test and carried the defect across without noticing. The remainder after the name must be a bare extension, carrying no dot of its own. The regression test is mutation-verified: removing the check fails it.

### Codex — Major: my "static superset" was not a superset

`timber_kit_resizer_allowed_types` runs *after* the capability intersection, so it can force a format **on**. A project that added one has derivatives `inputFormatMimes()` alone does not explain, and excluding that source would mis-resolve an ambiguous derivative — the exact wrong-placement class this change exists to remove.

The set is now the static list **widened** by that filter, and still never narrowed by the live backend. Fail-closed in both directions, with a test for each.

### Converged — flag-off was not byte-identical after all

Both reviewers, independently: the URL encoding ran unconditionally. A site whose `sanitize_file_name` filter leaves a space or a `+` would have seen its flag-off URLs change bytes on a version bump. Gated on the flag now, with a test.

Convergence again on the thing I was most confident about. Worth noting that "flag off is untouched" was my claim in three places.

### Fable — a guess by omission

A source whose *filename* failed the guard was dropped from the migrator's map entirely, so a sibling sharing its legacy name looked unique and the planner moved that sibling's derivative. Asymmetric with a refused *directory*, which already contributed a flat candidate. The candidate is kept now and refused at the point of use.

### Fable — documentation that contradicted itself

- ADR 0008's Decision section still specified the sanitize step its own amendment retracts. The file stated two contradictory specs, and only a reader who got to the amendment learned which binds.
- `StarterBase` carried a paragraph explaining why the name must be sanitized before `glob()` **directly above** the paragraph explaining why it is not.
- Three stacked docblocks above one method in `Resizer`, and `guardSourceDir` had lost its own — **the same misattachment I fixed once on #152, reintroduced by my own insertions.**

Also added where an operator will actually read it: the MIME exclusion, the case-insensitive filesystem pair, path-shape rules (CDN/nginx/`.htaccess`, and scripts that grep or rsync names with spaces), and the order of migrating versus changing a sanitize filter.

Six orphaned docblocks in `StarterBase` are left alone — they are on `main`, outside this PR.

### Correctness oracle re-run against the new naming

Cache emptied, 260 pages crawled, every scoped derivative compared to the source its own path names:

```
matches its own source : 598
MISMATCH               :  41   <- the known alpha artefact of the oracle
source file missing    :   0
```

**Zero missing sources, against 190 under the old sanitized naming.** That is the verbatim key round-tripping: a derivative's path now names a file that exists. The 41 all sit at ~0.457 in `220x220-top`; flattening alpha consistently on both sides drops a measured one to 0.0055 — same picture, diagnosed earlier in this PR.

parisek:
## Cross-checked on a second, deliberately different project (`pm-a`)

Sloneek is atypical in exactly the three places this change strains hardest, so it was validating the exception rather than the rule. `pm-a` inverts all three: **no WPML**, **911 attachments and not one of them in a directory** — every upload at the uploads root — and a 1214-file flat cache.

### The finding that reframes the whole PR

`pm-a` runs the same lossy `sanitize_file_name` filter as sloneek. It is **not a plugin, and not a site quirk.**

It is `StarterBase::clean_uploaded_filename()` — registered by *this package* whenever `$clean_image_filenames` is true, **which is the default**. It removes accents, lowercases, maps whitespace *and underscores* to hyphens, and strips the rest.

Right at upload, where it is a deliberate normalisation. Wrong as a cache key.

So the collision class is not one site's misfortune: it is the default for every consumer that has not turned the setting off. ADR 0008's amendment blamed "a plugin on the site this was tested against" and is corrected — the first version argued the collisions were theoretical because it went looking for a *third party* doing something unusual, and **the collapsing function ships in this repository.**

### The branch sloneek could never exercise

Everything on `pm-a` is a root upload, so the flag changes only the filename. The result:

| run | scanned | move | ambiguous | orphaned | conflicts | already in place |
|---|---|---|---|---|---|---|
| first | 1214 | **1214** | 0 | 0 | 0 | 0 |
| second | 1214 | 0 | 0 | 0 | 0 | **1214** |

All 1214 move; a second run classifies all 1214 as `already_in_place`. That is `d6ddbc7`'s already-migrated-root-upload branch — until now covered only by a unit test — proven on real files.

It also settles the stale docblock Fable caught. It told a site with year/month folders off that it "sees no change at all". `pm-a` is exactly that site, and **every one of its 1214 derivatives moves.** An operator who trusted that sentence would have orphaned the entire cache.

### Control run, so the version jump is not doing the work

`pm-a` sat on timber-kit `v1.32.0`. Installing this branch is a jump to `1.46-dev` with plenty of unrelated changes, so the flag was left **off** first: three pages rendered, **0 derivatives regenerated** on a 1214-file cache. The jump is inert for the cache; what follows is the flag.

After migrating with the flag on: pages render, **0 regenerated**, warm in 0.15s.

### Delete path where it actually fires

On sloneek every attachment carries ~5 WPML rows, so `attached_file_is_shared()` almost always keeps the files and the delete barely runs. `pm-a` is single-language, one row per file:

- `dtest.jpg` deleted → its derivative **gone**
- `dtest.png` — same stem, different extension, same (root) directory → **survived**

Under the flat key both were `dtest.avif`, so deleting one took the other. This is that collision at the uploads root, on a site where the guard does not mask the behaviour.

### Housekeeping

`pm-a` is restored: cache from its pre-test copy, `composer.json`/`.lock`/`Base.php` reverted, working tree clean, project stopped. It gained an imported database it did not have; nothing else changed.

Unrelated: local PHP is now 8.5, where `ReflectionMethod::setAccessible()` is deprecated, so the suite reports 12 deprecations. **The test suite on `main` reports the same 12** — pre-existing, fleet-wide, and invisible to CI, which runs 8.3 and 8.4.

Issue #147, opened 2026-08-26
perf: the options-page formatFields() is what is left of the context rebuild


parisek:
> **Rewritten 2026-08-27.** The original framing said `timber/context` costs 751 ms and proposed caching the options-page `formatFields()` to recover it. Both halves are now measured wrong. The old text is kept at the bottom.

## From which project

`sloneek` (redesign branch), re-measured after #154, #155 and #156.

## What is actually in the context, measured

The 751 ms figure came from a profiler and was an **inclusive** number on a filter. Inclusive time on a filter attributes every descendant to the filter, so it said almost nothing about the kit.

Re-measured by bracketing the filter with `microtime()` in a real web request — no profiler, no inflation — over ten runs, minimum values, with Redis warm:

| | |
|---|---:|
| whole request | 500 ms |
| `timber/context`, total | **145 ms** |
| … `parent::timber_context()` — the kit's own work | **1.7 ms** |
| … `Helpers::formatMenu()` × 5 menus | 87 ms → **fixed by #155** |
| … `Helpers::formatFields()` on the options page | **40 ms** ← this issue |
| … the project's own tail | ~15 ms |

Two corrections worth stating plainly:

- **`StarterBase::timber_context()` costs 1.7 ms.** There was never 751 ms of kit code in that filter.
- **The menu was the expensive half, not the options page**, and it is now cached. #155 took the front page from 410 ms to 362 ms.

## What is left

`Helpers::formatFields( <options page> )` costs **~40 ms** on the front page. Every downstream project makes the same call from `timber_context()`, so it is worth removing — but it is 40 ms, not 751, and that changes what is worth spending on it.

A second call in the same request costs 1.6 ms, so nothing here is repeated work. The 40 ms is the cold build: 977 stored rows walked through `fieldFormatter()`.

## What #155 already settled

The original text stopped here to ask how to invalidate. Most of that is now answered, because a menu is the same problem: formatted output cached across requests, with the same dynamic surfaces.

Reusable as-is:

- **`CacheSignature::shared()`** — site, language, roles, and a content version from `wp_cache_get_last_changed()` over `posts` and `terms`. This answers the original question 1 (link fields embed permalinks): a slug change bumps the `posts` counter and the old key goes unreachable. No hook needed.
[...]

Issue #131, opened 2026-08-18
perf(wpml): opt-in guard for ACFML's front-end ACF field-definition translation


parisek:
## Problem

ACFML translates ACF field *definitions* — labels, instructions, placeholders,
prepend/append, choices, message — on every request that loads a field group.
`ACFML\Strings\FieldHooks` implements `IWPML_Frontend_Action`, so it registers on
the front end with no `is_admin()` guard, and hooks `acf/load_field`. Every field
then reaches `Config::getForField()`:

```php
Fns::filter( Relation::propEq( 'namespace', 'field' ), self::DATA )
```

An unmemoised linear scan over a static array, run through WPML's functional
library, allocating a closure per element. On a plain page view none of those
strings is rendered, so the whole walk is wasted.

## Measured (sloneek, 117 field groups / 972 top-level fields)

| | time |
| --- | --- |
| baseline | 1.09–1.13 s |
| group-level translation off | 1.04–1.14 s (no change) |
| **field-level translation off** | **0.29 s** |

A PHP-FPM slowlog profile over 3973 slow requests put **71 %** of deepest-frame
samples inside `wpml/fp` + `wpml/collect`.

Staging, same box, alternating the file to hold load constant:

| round | without | with |
| --- | --- | --- |
| 1 | 9.70 s | 4.17 s |
| 2 | 8.97 s | 4.28 s |
| 3 | 10.58 s | 3.99 s |

Rendered HTML is byte-identical once per-render `uniqueId()` values are
normalised.

## Why this is not #107

[#107](https://github.com/parisek/timber-kit/pull/107) was closed unmerged, on
the reasoning that a shared workaround would outlive the bug it worked around.
That reasoning was right there and does not carry here:

| | #107 (`acf/load_reference` leak) | this |
| --- | --- | --- |
| who pays | one caller outside WPML's own plugins | every front-end render |
| fixed upstream | **yes**, acfml 3.0-b.1 | **no** |
| workaround lifetime | would outlive the bug | bug has no end date |

Verified rather than assumed: on `proficiohub` (acfml 3.0-b.1, sitepress
5.0.0-b.1, wpml-string-translation 5.0.0-beta.1), `Strings/FieldHooks.php`,
`Config.php`, `Translator.php`, `Traversable/Entity.php`,
`Traversable/FieldGroup.php`, `Transformer/Translate.php` and `Package.php` are
byte-identical to 2.2.4, and the bundled `wpml/fp` library has identical MD5s.
[...]

PR #154, opened 2026-08-27
feat(kit): run the backend and field-group probes once per request

parisek:
## From which project

`sloneek` (redesign branch), profiling an uncached front-page render locally with xhprof. Companion to #147 and #131, which own the two larger items in the same profile; this is the self-contained part.

## The finding

Two capability probes ran once per object instead of once per request. **Both already had a memo** — stored on the instance the caller throws away, so neither ever answered anything. `Resizer`'s docblock said `Memoized per request`, which is the sentence that kept anyone from checking.

`acf_get_field_groups()` is the more interesting one: **ACF caches nothing there.** Repeating the *identical* screen costs the same 8-10 ms as a new one, because every call walks each registered field group and evaluates its location rules. `formatFields()` asks once per nav menu item, so a 90-item menu paid for 90 answers.

## Measured

By patching the site and re-measuring, not by scaling the profile. Fastest of twelve warm requests, **with no configuration of any kind**:

| | Imagick probes | `acf_get_field_groups()` | page |
|---|---:|---:|---:|
| before | 320 | 109 | 671 ms |
| after | 1 | 11 | **561 ms** |

**110 ms, 16 %**, reproduced across three rounds at 101-112 ms. The saving scales with images and menu items rather than with the page.

Output is byte-identical once the random `uniqueId()` values are normalized. That check needed a control pair of runs of the *unchanged* code first, because the page is not deterministic without one — the first comparison showed differences that had nothing to do with this change.

## How the nav-menu sharing stays correct

Sharing one field-group answer across every item of a menu is where most of the saving is. It is safe while ACF's own location types are the only things that see the screen: `ACF_Location_Nav_Menu_Item::match()` reads the item id only to confirm the key is set, then matches on `nav_menu`.

It is **not** safe in general. `acf_register_location_type()` is public API, ACF hands the whole screen to every registered type, and a *"show this group on this one menu item"* rule — a mega-menu wanting extra fields on one item — is a legitimate thing to write. Sharing would then serve the first item's groups to every item, with no error and no log.

So the kit checks rather than guesses. Before sharing, **every registered location type must resolve to a class file inside `ACF_PATH`**. A site that registers its own gets per-item lookups back automatically, with nothing to configure and no notice to read. Anything unverifiable — no `ACF_PATH`, an empty registry, a class with no file — counts as unsafe. The check costs 0.003 ms and is memoized per request.

It reads the location-type registry rather than scanning field-group rules, because `acf_get_field_groups()` with no arguments costs 27 ms on a cold store — a fifth of the saving, spent protecting it.

`timber_kit_share_nav_menu_item_field_groups` overrides the verdict either way: force it off for an `acf/location/rule_match` callback that reads the item id (the one path the check cannot see; ACFML's reads `post_id`, `lang`, `page_parent` and is fine), or force it on for a custom type known not to read it.

**This whole mechanism exists because of the Codex review below.** The first version normalized unconditionally.

## Not done here, deliberately

- **Neither memo moved into the object cache.** They remove repeated work inside one render, not the first call of each request. Persisting them is a separate decision whose cost is invalidation, not implementation: ACF field groups load from theme JSON files and no hook fires when a deploy changes one. A persistent cache would also still want a static in front of it, because `wp_cache_get()` is not free. The place where that trade pays off is #147, where one filter costs 751 ms.
- **No flush on late `acf_add_local_field_group()`** — ACF exposes no action for it. The README states the limit rather than pretending to cover it.
- **`decodableFormats()` keeps its per-instance memo**, so `timber_kit_resizer_allowed_types` still runs per instance. Only the raw backend capability is shared, and that is the filter's input, not its output.

## Type

`feat`, not `perf`, because public methods and a filter are new and RELEASING.md maps a new public symbol to MINOR. The intent is performance.

## One thing worth reading

Ten unrelated resizer tests failed at first, and the cause was **the new test file existing**, not the memo. A `class X extends Resizer` at the top level of a `*Test.php` file loads `Resizer` while PHPUnit is still collecting files — before Patchwork instruments it — and `Patchwork\redefine( 'file_exists', … )` then stops reaching the copy those tests exercise. The stub subclasses now live in `tests/Fixtures/`, autoloaded lazily. Bisecting was the only way to find it; the failure pointed squarely at the production change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,53 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Changed
+
+- Two capability probes that ran once per object now run once per request.
+  Both had a memo already; both stored it on the instance the caller throws
+  away, so it never answered anything.
+
+  `Resizer::probeBackendFormats()` built a `new Imagick()` and asked it for the
+  format list. One `|resizer` call builds one Resizer, so a page that resizes
+  320 images probed 320 times — for a list that is a property of the
+  ImageMagick build and cannot change while the process runs. The memo moves to
+  a static; `Resizer::flushBackendFormats()` drops it.
+
+  `Helpers::getFieldObjectsByScreen()` called `acf_get_field_groups($screen)`
+  per nav menu item. ACF caches nothing here: every call walks each registered
+  field group and evaluates its location rules, measured at 8-10 ms against 96
+  groups whether or not the screen repeats. Answers are now memoized per
+  screen, and `Helpers::flushFieldGroups()` drops them.
+
+  The key normalizes `nav_menu_item` to a presence marker, which is what makes
+  the memo hit at all. `ACF_Location_Nav_Menu_Item::match()` reads that key only
+  to confirm it is set and then hands the decision to the `nav_menu` location
+  type, so the item id cannot change which groups match — every item of one menu
+  shares an answer. The key also carries the blog id and the language, because
+  field groups are registered per site and ACFML translates a group's own
+  strings.
+
+  Measured on the sloneek front page by patching the site and re-measuring, not
+  by scaling the profile. Median of eight requests, warm:
+
+| | before | after |
+  | --- | ---: | ---: |
+  | `Imagick::queryFormats()` calls | 320 | 1 |
+  | `acf_get_field_groups()` calls | 109 | 24 |
+  | page | 644 ms | 548 ms |
+
+  That is **96 ms, about 15 %**. The rendered HTML is byte-identical once the
+  random `uniqueId()` values are normalized — checked against a control pair of
+  runs of the unchanged code, because the page is not deterministic without one.
+
+  The saving scales with images and menu items, not with the page: the same
+  measurement on an archive with fewer images gave 6 %.
+
+  A static memo is request-scoped for a web request and **not** for WP-CLI or a
+  persistent worker, where one process can register field groups and then read
+  them back. Both flush methods are public for those callers; `StarterBase`
+  wires `flushFieldGroups()` to `acf/update_field_group`.
+
 ### Fixed
 
 - `StarterBase::cleanup_cached_images()` deleted resizer derivatives that other
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   two in-process memos below. A web request never needs either; WP-CLI
   commands, persistent workers and tests do, because a static outlives the unit
   of work that filled it.
+- `timber_kit_share_nav_menu_item_field_groups` — opt in to letting every item
+  of one menu share a field-group answer. **Off by default, and the default is
+  the safe one.** See below for the contract it asks you to accept.
 
 ### Changed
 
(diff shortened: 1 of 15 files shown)


parisek:
## Agent review record — 2026-08-27

Two reviewers, run in parallel and blind to each other, on different lenses. Neither is a formal approval.

| | Tool / model | Lens |
|---|---|---|
| 1 | Codex (`codex:rescue`) | Correctness of the cache key |
| 2 | Fable (via `Agent`) | Lifecycle and boundaries of the new static state |

**They converged on nothing.** Every finding below was raised by exactly one of them — which describes the lenses, not the findings. Splitting the lenses is what made that expected rather than surprising.

### Fixed

| Finding | From | What changed |
|---|---|---|
| `nav_menu_item` normalization is not extension-safe. `acf_register_location_type()` is public API and ACF hands the whole screen to every matcher and to `acf/location/rule_match`, so a custom matcher may legitimately answer per item — and a screen whose value is literally `'*'` collides. | Codex | Gated behind `timber_kit_share_nav_menu_item_field_groups`, **default off**. Costs 58 ms of the 112 ms saving by default; the rest is opt-in. `e38b2dc` |
| `wp_json_encode()` can return `false`; casting it to a string collapsed every unencodable screen onto one key. | Codex | Such a screen is no longer memoized at all. Test added. `e38b2dc` |
| The `Resizer` memo was one shared static. A subclass overriding `probeBackendFormatsUncached()` filled the entry ordinary instances read — the test fixture demonstrates it. | Codex | Keyed by `static::class`, with a test using two differently-stubbed subclasses. `e38b2dc` |
| `acf/delete_field_group`, `trash` and `untrash` were not wired. | Codex | All four now wired. **My earlier commit claimed no such actions exist — that was wrong.** ACF fires them dynamically as `acf/{$verb}_{$hook_name}`, so a literal grep misses them, and `local-json.php:51-54` listens to all four. `e38b2dc` |
| Nothing asserted the flush hook was registered, so deleting the line changed no test. | Fable | Test added, verified to fail without the line. `6d7beb9` |
| The two new public methods appeared only inside the `### Changed` narrative. | Fable | Moved to `### Added`. `6d7beb9` |

### Rejected

| Finding | From | Why not |
|---|---|---|
| List in the README the concrete callers that do not boot `StarterBase`. | Fable | Such a list goes stale faster than anyone reads it, and a stale list is worse than the general rule. The rule stays. |
| Enforce test isolation statically (process isolation, or a CI grep for classes not extending the two `TestCase`s). | Fable | Fable verified the suite is correct today and called it future fragility. Both proposed guards cost more than the defect; this is a pin to add the first time it actually bites. |
| Flush on late `acf_add_local_field_group()`. | Codex | ACF exposes no action for it. The README states the limit instead of implying coverage. |
| Make the whole memo opt-in rather than changing helper semantics. | Codex | The screen-keyed memo is correct for any screen without further assumptions; only the nav-menu-item sharing needed a gate, and that is what it got. Making the rest opt-in would leave a known-safe saving switched off by default. |

**Codex cited `local-json.php:35-38` for the delete/trash/untrash hooks; those lines are a property docblock.** The substance was right — the hooks exist and that file does listen to all four, at lines 51-54 — but the citation pointed at the wrong thing. Its conclusion was checked against the source before acting on it, which is how the misfire surfaced.

Codex also framed the ACF-impurity finding (late registration, mutable filters) as **blocking**. It is real and is now documented, but for a web request field groups register on `init` before rendering, so the exposure is long-running processes — which the public flush already addresses. Treated as documentation, not a blocker.

### Verification

`composer check` green after every step: 1768 unit tests, 20 property tests, PHPStan level 8, ADR index. Each new test was verified to fail without the code it covers.

parisek:
## Follow-up: the sharing is on by default again, and now it verifies itself

The Codex finding above was right, and the first fix over-corrected: putting the sharing behind an opt-in filter left **58 ms of the 110 ms saving switched off on every site that did not know to look for it**. Off-by-default is the wrong trade for something that is correct on any ordinary ACF install.

`be22b83` makes it default-on **and** removes the assumption instead of documenting it. Before sharing, every registered location type must resolve to a class file inside `ACF_PATH`. That is exactly the risk Codex named — `acf_register_location_type()` is public API — and it is structurally detectable, so nobody has to know about it:

| Site | What happens | Configuration needed |
|---|---|---|
| Ordinary ACF (with or without ACFML) | shares, 109 lookups → 11 | none |
| Registers a custom location type | per-item lookups, automatically | none |
| Custom type known not to read the item id | filter forces sharing on | one line |
| `acf/location/rule_match` callback reading the item id | filter forces sharing off | one line |

Unverifiable counts as unsafe: no `ACF_PATH`, an empty registry, a class with no file, a non-object in the registry. Five tests cover the verdicts, using a fixture in a directory the test points `ACF_PATH` at — the same mechanism the production check uses, not a parallel one.

The check reads the location-type registry rather than scanning field-group rules. `acf_get_field_groups()` with no arguments costs 27 ms on a cold store, which would have spent a fifth of the saving protecting it; the registry read costs 0.003 ms and is memoized per request.

Re-measured with **no configuration at all**: 320 Imagick probes → 1, 109 lookups → 11, page 671 ms → 561 ms, output byte-identical against a control pair.

The residual gap is unchanged and stated in the README: a location-match filter reading `nav_menu_item` is not detectable this way. Writing menu-item-specific logic there rather than registering a location type is contrived, and ACFML's callback reads `post_id`, `lang` and `page_parent`.

PR #155, opened 2026-08-27
feat(helpers): cache the page-independent half of a formatted menu

parisek:
## From which project

`sloneek` (redesign branch). Follow-up to #154 and #156: with those merged, `Helpers::formatMenu()` is the single largest item left in the render.

## Measured

Front page, real Redis object cache, two back-to-back runs of 15 requests each:

| | before (v1.43.0) | after |
| --- | ---: | ---: |
| front page, min | 402 ms | **344–350 ms** |
| `/blog/`, min | 907 ms | **836–857 ms** |

That is ~53 ms, **13 % of the request**. The isolated cost of the work removed is ~87 ms across five menus (72.5 ms on the 68-item main menu, 14.4 ms on four footer menus), measured by bracketing the call in a real web request. The gap between 87 and 53 is the half that is deliberately *not* cached — see below.

Rendered HTML is byte-identical once non-deterministic ids are normalized, with a valid control pair on each side.

## What is stored, and what is not

The ACF fields on each menu item and on the menu itself are the same on every URL. `is_active` and `in_active_trail` are not, and the walk recomputes them on every request including a cache hit.

Caching the whole formatted menu instead would freeze the highlight onto whatever page filled the entry — wrong on every page but one, and invisible to any status check. The split also keeps this to one entry per menu rather than one per page.

## The gate proves purity before the work, not after it

This is the part that changed since the first draft, and it is the point of the PR.

The earlier gate watched the render and asked whether formatting had changed anything. **That is not a proof of purity.** An unregistered `[foo]` comes back byte-identical, reads as static, and the literal `[foo]` is stored. The day a plugin registers that shortcode, every page serves the frozen source text instead of its output — with nothing to log and nothing to notice. A reviewer broke the observational gate with a three-line shortcode, and the conclusion was that observing one render cannot prove purity at all.

So each of the three dynamic surfaces in `fieldFormatter()` is now decided from stored data, before the build:

| Surface | Static proof |
| --- | --- |
| `do_shortcode()` on `wysiwyg`/`textarea` | no `[` in any of the item's raw meta strings |
| `field_formatter_{$type}` filters | none registered — no static proof of a callback's purity exists, so one ends it for every menu |
| CF7 / WPForms `post_object` | none available: the markup carries a nonce, so this one surface stays counted during the build |

The meta read is free in practice — `wp_get_nav_menu_items()` has already primed it — and only runs when the walk can actually store.

The check is deliberately blunt. A bracket in a plain-text field is not a shortcode and the menu is refused anyway: **a false refusal costs one uncached menu, a false accept is wrong on every page.**

The filter refusal is a default, not a verdict. It goes through `timber_kit_cache_menu_fields` with everything else, so a project that knows its own formatter is pure can say so.

## Rejected

- **Fingerprinting the registered `field_formatter_*` tags into the cache key** instead of refusing. It catches a callback appearing or disappearing, but it cannot make a page-dependent callback safe — which is the actual risk.
- **`<img` in the gate.** `fieldFormatter()` never runs `wp_filter_content_tags`, so an image tag is not a dynamic surface here. Including it would refuse menus for no reason.
- **Memoizing `CacheSignature::shared()`.** Two independent reviews found the same mid-request staleness hole from different directions: `switch_to_blog()`, a user switch and a save in a long-running process each move an input, and a memo would key the second site's menu under the first site's name.

## Invalidation

`CacheSignature::shared()` carries site, language, the current user's roles and a content version from `wp_cache_get_last_changed()` over `posts` and `terms`. WordPress bumps those counters itself, so a saved post or an edited term makes the old key unreachable rather than stale. Nothing has to be flushed and nothing can be forgotten.

The TTL (`timber_kit_menu_fields_ttl`, 12 h) is not about staleness — it bounds the generations each content change orphans, which nothing else deletes.

Requires a persistent object cache; without one the path is skipped rather than paying for a read and a write that always miss.

## Tests

Each gate was verified load-bearing by removing it and confirming the failure:

- disabling the stored-value check fails `test_shortcode_rendered_output_is_never_stored` and `test_an_unregistered_shortcode_is_refused_although_nothing_changed`
- disabling the filter check fails `test_a_registered_field_formatter_filter_refuses_every_menu`

1811 tests green, PHPStan clean.

https://claude.ai/code/session_01JMYy6JHLf4mU4H4Hd47spb


parisek:
## Agent review, round 2 — 2026-08-27

Same two reviewers, same lenses, re-run after `bede0ee`.

### Round 1 fixes: both reviewers confirm they hold

| Finding | Raised by | Verdict now |
|---|---|---|
| `CacheSignature` memoized inputs that move mid-request | both | Fixed beyond doubt. Nothing is memoized; mid-request site and user switches are tested. |
| open/close not exception-safe, re-entrancy lost the write | both | Fixed for exception, same-menu re-entrancy, nested different menu, repeated calls. |
| the cache belonged in its own component | Fable | `MenuFieldsCache`; boundary confirmed correct. |

### The new finding, and it is the one that matters

**Codex broke the dynamism detector with a counterexample I cannot answer:**

```php
add_shortcode('contextual', fn() => is_singular('product') ? 'Buy now' : '[contextual]');
```

Out of context the shortcode returns *its own source text*. The value does not change, the detector reads it as static, and the literal `[contextual]` is stored. On a product page the cache hit bypasses `do_shortcode()` entirely and renders the literal.

The same shape defeats the `field_formatter_*` comparison: a callback that varies by post, user or capability can look inert on the seeding request.

**The general point is right and is not fixable by a better detector: observing one render cannot prove purity across renders.** Round 1's detector was too blunt (it rejected the largest menu on the measurement site); round 2's is too permissive. There is no middle setting, because the property being tested is not observable from one sample.

Codex also found that a nested *incomplete* close is forgotten — `close($id, false)` at depth > 1 only decrements, so an inner walk that threw and was caught still allows the outer write. Real, and separate from the above.

### Why this is not being patched

The obvious fix is Codex's: cache raw ACF values and format on every request. **Measured on the front page, that would still capture most of the win** — the lookup is 56-102 ms of the 64-118 ms per-item cost, so formatting is only ~12 % of it.

But there are no raw values at this boundary. `getFieldObjectsByScreen()` fills `$field['value']` from `get_field()`, and ACF's own `format_value` runs `the_content` filters for wysiwyg — shortcodes included. Caching "before our formatting" still caches shortcode output. The safe boundary is `get_field( …, false )` plus live `acf_format_value()`, which changes what every downstream consumer of `formatFields()` receives.

That is a larger and riskier change than the 12 % of one page it would buy, on top of #154 which already takes the larger bite with no staleness surface at all.

### Recommendation: do not merge; park it

Marked draft. The work is not wasted — `CacheSignature`, the depth-counted protocol, the storability policy and the tests are all sound, and they are what a later attempt at the deeper boundary would build on.

### Fable's smaller findings, for the record

- `Helpers::$dynamic_format_count` is never reset in tests. Harmless today because only a delta is read, but a trap for the first test that reads it absolutely.
- No test covers menu A whose formatting re-enters `formatMenu()` for menu B; the mechanism handles it, nothing pins it.
- The CF7/WPForms branches increment unconditionally while everything else compares — correct in practice, inconsistent with the stated doctrine, and unexplained in the code.

All three would be worth fixing if this were shipping. They are recorded rather than done.

parisek:
## Agent review, 2026-08-27 — Codex (gpt-5.3-codex) and Fable, run blind to each other

Two reviewers, different lenses: Codex on cache soundness and the purity gate, Fable on downstream blast radius, observability and API surface. Neither saw the other's output.

### They converged on one thing, from opposite directions

**The ACF value-load layer.** Codex reached it directly: the gate read raw post meta, but values arrive through `get_field()`, and ACF can supply one no meta row holds. Fable reached the same layer from the other end, arguing the roles-only audience signature cannot model two same-role visitors getting different values.

That convergence is what makes it credible, and it was **verified against the installed ACF source before acting** — `acf-value-functions.php::acf_get_value()` really does apply `acf/pre_load_value` first, fall back to `default_value` when no row exists, and then run `acf/load_value`. Group and clone sub-fields take their own defaults the same way.

So a menu whose meta was provably bracket-free could still put `[contextual]` through `do_shortcode()`, and this cache would have stored one request's rendered output onto every page. **My design, my error** — the previous commit's gate was unsound.

### Fixed

| Finding | Reviewer | What changed |
| --- | --- | --- |
| Gate reads the wrong store (**critical**) | Codex | The check moved to the `do_shortcode()` call site and inspects its **input**. Same property — decided before the dynamic call, never from whether output differed — but on the value ACF actually produced. Less code, one fewer meta read per item. |
| No way to purge a wrong entry now | Fable | `MenuFieldsCache::flushStored()` / `Helpers::flushStoredMenuFields()`, group-scoped, guarded by `wp_cache_supports('flush_group')`. Fable correctly pointed out `BlockRenderer.php:396` already did exactly this — the precedent was in the same codebase and unused here. Without it the only lever was `wp_cache_flush()`: every group, every site on shared infrastructure. |
| The counter-test | Codex | `test_a_shortcode_acf_supplies_but_no_meta_row_holds_is_refused` keeps meta **empty** on purpose. It fails against the previous commit. |

Each gate re-verified load-bearing by removal: disabling the input check fails exactly the three shortcode tests and nothing else.

### Rejected

- **Refusing when any `acf/load_value` callback is registered**, mirroring the `field_formatter_*` treatment. ACFML registers one on every WPML site, and its callback converts link targets by language — and language is already in the key, so it is a function of inputs the key carries. A blanket refusal would disable the feature precisely where it was measured, to guard against a callback shape neither reviewer found in practice.
- **Fingerprinting registered formatter tags into the key** rather than refusing. Catches a callback appearing or disappearing; cannot make a page-dependent callback safe, which is the actual risk.

Fable's blast-radius list named membership and personalization plugins that filter `wp_get_nav_menu_items` as leak vectors. **They are not.** The payload is keyed per menu item id and holds only that item's ACF fields; a plugin that removes items changes which slots are asked for, not what a slot contains. The mechanism Fable identified is real, but its examples were the wrong ones — the actual vector is the ACF value layer Codex found. Worth knowing when calibrating the rest of that report.

Codex could not reach the GitHub API and reviewed from the branch diff and commit message rather than the PR description.

### Open, and needing an owner's decision rather than a commit

1. **ACF field-group JSON changes bump neither `posts` nor `terms` last-changed**, so a deploy that edits a `default_value` or a return format leaves entries reachable for up to the TTL. Codex rated this High. A config version in the signature would fix it, at a per-request cost.
2. **An `acf/load_value` callback reading the current user** is undetectable by any static means. Fable's argument that this makes on-by-default the wrong call is coherent, and is the strongest case for shipping the whole path opt-in for one release.
3. **No observability.** No hit/miss signal, and `timber_kit_cache_menu_fields` receives a bool and a menu id but never the reason the default said no.

Both reviewers were right that these matter. I have not decided them here because (1) and (2) trade correctness against the reach the feature was built for, which is the owner's call, not a reviewer's and not mine.

parisek:
## The three open items are now four commits

Each is independent and can be rejected on its own.

| Commit | What |
| --- | --- |
| `ba657df` | field-config version in the key |
| `7e19936` | ACF value-load callbacks judged by where they are defined |
| `5a99dd7` | the refusal reason is reported instead of discarded |
| `7f87923` | fixes `ba657df`, which versioned nothing — see below |

### 1. Field definitions change without touching content

The key carries a hash of the `modified` stamps of the groups the menu reads. ACF's local-JSON loader takes `modified` from the file, so an admin edit moves it and so does a deploy.

**The first attempt at this shipped broken and the tests did not catch it.** It asked the `nav_menu` screen only, reasoning that item groups are located by the same rule. `ACF_Location_Nav_Menu_Item::match()` does delegate to `nav_menu` — but only after confirming `nav_menu_item` is **set**. With that key absent it matches no item group at all. Measured: the `nav_menu` screen returns zero groups on the reference site; the item screen returns the one holding every menu field. So the token was the constant `nogroups` on every menu.

It was visible only because the Redis key is printable:

```
menu-fields-478|nogroups|b1|len|aanon|p0.795…|t0.832…     ← before
menu-fields-478|2b141291ff39|b1|len|aanon|p0.795…|t0.832… ← after
```

The unit tests passed against both implementations, because the stub returned the same group for any screen. `7f87923` adds one that reproduces the real asymmetry and fails against `ba657df`.

`nogroups` is still the right answer for a menu carrying no ACF fields — four of the five here.

### 2. `acf/load_value` — judged by origin, not by presence

Refusing on presence would have been wrong and measurably so. Both hooks are occupied on the reference site: ACF's own `_acf_apply_hook_variations()`, which is what makes `acf/load_value/type=…` fire at all, and ACFML's link converter. A presence test switches the cache off exactly where it was measured.

So callbacks are judged by where they are defined — the same test `navMenuItemSharingIsSafe()` already applies to location types. ACF is trusted because it is the mechanism, not a policy. ACFML because it varies on the language, and the language is already in the key. `timber_kit_trusted_value_load_roots` lets a project vouch for its own.

**The variation tags are scanned too, and that is load-bearing.** The reference site has fourteen of them live and not one is named `acf/load_value`. A scan of the two literal tags would report "nothing registered" and store.

Verified against the live site after the change — all five menus still cache:

```
array ( 478 => 'cache', 482 => 'cache', 483 => 'cache', 484 => 'cache', 75 => 'cache' )
```

### 3. Observability

`Helpers::menuCacheDecisions()` returns the decision per menu. The reason also reaches `timber_kit_cache_menu_fields` as a third argument, which is what changes what a project can express: a boolean only says "cache this whatever you found", a reason lets an author say "that formatter is mine and it is pure" while leaving every other refusal standing.

Additive — a callback registered with two accepted args keeps receiving two.

### Not done, deliberately

- **The `acf/update_field_group` flush hook** proposed alongside item 1. Redundant: `flushFieldGroups()` already drops the memo on all four ACF verbs, so the next version read sees the new `modified` by itself.
- **Making the whole path opt-in.** The two gates above close the surfaces the reviews actually found, and the owner's position is that a cache useful to every site should not need a filter to turn on. A site with an unrecognised value-load callback now gets the refusal automatically instead.

### Measured, after all five commits

| | before (v1.43.0) | after |
| --- | ---: | ---: |
| front page, min | 410 ms | **371 ms** |
| median | 427 ms | 400 ms |

Rendered HTML byte-identical, valid control pair. 1823 tests, PHPStan clean, CI green. Every gate re-verified load-bearing by removal.

PR #157, opened 2026-08-27
feat(seo): self-referencing canonical on paginated routes

parisek:
## The defect

A listing rendered by a block on an ordinary page is singular as far as an SEO plugin is concerned, so the plugin resolves the canonical to `get_permalink()` and every page of the listing claims to be page one.

Measured on sloneek.com on 2026-08-27: `/blog/page/2/` and `/blog/page/10/` both emitted `canonical https://www.sloneek.com/blog/`, as did `/case-studies/page/2/` and `/ebooks/page/2/` for their own routes. Nine tenths of each listing was telling crawlers it duplicated page one.

## Why an adapter layer, not a snippet

The project carrying this defect is migrating Yoast → All in One SEO. A fix written against `wpseo_canonical` alone is lost the day the switch completes; one written only against `aioseo_canonical_url` cannot ship before it. `Seo\Canonical::register()` hangs the same filter callback on whichever plugin `Seo\Plugin::active()` finds loaded — `Seo\Yoast::register()` on `wpseo_canonical`, `Seo\Aioseo::register()` on `aioseo_canonical_url` — so the fix survives the migration instead of needing to be redone on the other side of it.

## The `function_exists` testability constraint

Brain\Monkey stubs a WordPress function by defining the real global function, and that definition outlives the test that made it — so once any test in a run has stubbed `function_exists( 'aioseo' )` or `defined( 'WPSEO_VERSION' )`, the false branch of that check becomes untestable for the rest of the run. The package's own `AGENTS.md` already documents this limit.

`Plugin::detect()` answers the pattern this leaves testable: it is pure, takes `array{yoast: bool, aioseo: bool}`, and is exhaustively covered, including the "both loaded" case (AIOSEO wins, because it is the migration target — a half-finished migration should behave like the destination, not like the plugin being removed). `Plugin::active()` is the one place that actually calls `function_exists`/`class_exists`/`defined` — kept to one statement per plugin, small enough to verify by reading rather than by test.

## The opt-in flag

`StarterBase::$seo_canonical_pagination`, default `false`, gates `registerSeoHooks()`. A site whose listings are real post-type archives already gets correct pagination canonicals from its SEO plugin and needs nothing from this — the flag exists so this only activates for sites like sloneek, whose listings are blocks on ordinary pages.

## What Yoast deliberately does not implement

`Seo\Yoast` wires the canonical filter only. Yoast is the plugin being migrated away from; writing its title and og:image adapters would be work invested in a dependency on its way out. `README.md` marks both as `n/a — plugin retiring`, a decision rather than a gap.

## The blind rule, and the two rejected alternatives

`Canonical::filter()` applies to any singular request past page 1, with no check for which block is on the page. Two alternatives were rejected:

- **`has_block()` detection** — a registry of paginating blocks that drifts the moment a second one is written, and cannot see a block nested inside a pattern or a reusable block.
- **Letting the block announce itself at render** — impossible by ordering: `wp_head` (where the canonical filter fires) completes before block content renders.

## Known follow-up

Yoast is now detected by two symbol sets: `Plugin::active()` uses `WPSEO_VERSION`/`WPSEO_Options`, while `WarmupSitemap::isYoastSitemapActive()` keeps `WPSEO_VERSION`/`WPSEO_Sitemaps`. That's deliberate — this PR scoped the consolidation to the canonical-pagination decision, not to `WarmupSitemap`'s own loaded-check — but it leaves two definitions of "Yoast is here" in the package, worth unifying later.

## Verification

`composer check` green: 1822 unit tests, 20 property tests, PHPStan level 8 clean, ADR index OK. Hygiene checks clean. `SeoBoundaryTest` was proven to fail when violated (not merely to pass), keeping plugin names inside this namespace.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,7 @@ New behavior that changes rendered output, admin behavior, or anything a consume
 - **Pattern.** Declare `protected bool $feature_name = false;` on `StarterBase` (grouped with related flags, docblock stating *what it changes* and *why it's opt-in*), then gate the wiring inside the relevant `registerXxxHooks()` method: `if ( $this->feature_name ) { add_action( … ); }`. Cover both branches (off → not wired, on → wired) in the matching `RegisterXxxHooksTest`. Examples: `$security_headers`, `$wpml_block_override`.
 - **Breaking changes are allowed — but only this way.** A breaking or behavior-changing change may land *provided* it's behind such a flag and defaults to off, so existing consumers are unaffected until they flip it. No silent behavior changes on upgrade.
 - **Opinionated defaults live downstream.** The `wordpress-base` project template enables these flags (`true`) in its own `Base` — that's where Porta Design's best-practice config is expressed, not in the library defaults.
+- **Approved exception:** `$seo_canonical_pagination` defaults `true`, not `false`. The owner reviewed it explicitly because `Pagination::append()` is idempotent and reads the site's own pagination base — on a site whose listings are real post-type archives the filter re-derives the identical canonical, so flipping the default changes nothing there. This is a named exception to the rule above, not a case of the rule being skipped.
 
 ## Architecture decisions (ADRs)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- `Seo\Canonical` — a self-referencing canonical on paginated routes, for
+  whichever of Yoast or All in One SEO is running. A listing rendered by a
+  block on an ordinary page is singular as far as an SEO plugin is concerned,
+  so it resolves the canonical to `get_permalink()` and every page claims to
+  be page one. Measured on one site: `/blog/page/2/` through `/blog/page/10/`
+  all pointed at `/blog/`, as did the same routes on two further listings.
+  Behind `$seo_canonical_pagination`, default `false` — a site whose listings
+  are real post-type archives already gets this from its plugin.
+
 ## [1.43.0] - 2026-08-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   so it resolves the canonical to `get_permalink()` and every page claims to
   be page one. Measured on one site: `/blog/page/2/` through `/blog/page/10/`
   all pointed at `/blog/`, as did the same routes on two further listings.
-  Behind `$seo_canonical_pagination`, default `false` — a site whose listings
-  are real post-type archives already gets this from its plugin.
+  Behind `$seo_canonical_pagination`, default `true` — see Changed below.
+  Reads the site's own `$wp_rewrite->pagination_base` rather than assuming
+  `page`, and leaves a manual canonical alone whenever it points somewhere
+  other than the current request.
+
+### Changed
+
+- `$seo_canonical_pagination` defaults `true`, not `false` — an approved
+  exception to this package's default-off rule for new behaviour (see
+  `AGENTS.md` § Feature flags & breaking changes). Safe on: the filter is
+  idempotent and a no-op on a site whose listings are real post-type
+  archives, since re-deriving the canonical from itself reproduces the
+  identical string.
 
 ## [1.43.0] - 2026-08-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   `page`, and leaves a manual canonical alone whenever it points somewhere
   other than the current request.
 
+### Fixed
+
+- `Seo\Canonical::filter()` was a silent no-op on every WPML + Yoast install.
+  WPML's Yoast glue (`wp-seo-multilingual`) registers its own
+  `wpseo_canonical` callback at the same filter priority and, on a tie, runs
+  first — handing this method a whole-string URL-encoded canonical
+  (`https%3A%2F%2F...`) instead of a real URL. `parse_url()` found no host in
+  that, so the current-request comparison always failed and every paginated
+  canonical passed through untouched, on every one of the five languages
+  tested. Now decodes once when the raw value has no host, before comparing
+  or appending the pagination segment; a value that still has no host after
+  one decode is treated as unusable, same as before.
+
 ### Changed
 
 - `$seo_canonical_pagination` defaults `true`, not `false` — an approved
--- a/README.md
+++ b/README.md
@@ -329,6 +329,47 @@ the database** (majority vote, tie-break toward core tables; `--collate=`
 overrides). Tables with COMPACT/REDUNDANT row formats and long indexed
 columns are flagged (767-byte index-prefix limit) and require `--force`.
 
+### Seo
+
+Everything the kit knows about an SEO plugin lives under `src/Seo/`
+(`Plugin`, `Canonical`, `PagedRequest`, `Pagination`, `Yoast`, `Aioseo`) —
+enforced by a boundary test, `tests/Unit/Architecture/SeoBoundaryTest.php`, the
+same way `src/Breeze/` is enforced by `BreezeBoundaryTest`.
+
+`Plugin::detect()` picks one plugin when more than one is loaded (AIOSEO wins
+— it is the migration target) and is used by both capabilities below, so the
+two never disagree about which plugin is running.
+
+| Capability | Yoast | AIOSEO |
+| --- | --- | --- |
+| canonical | ✅ | ✅ |
+| title | n/a — plugin retiring | not yet |
+| og:image | n/a — plugin retiring | ✅ (`SocialImageBridge`) |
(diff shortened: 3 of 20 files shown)


parisek:
Five commits added since the first review pass, from owner feedback. Two are defect fixes, one flips the default, and one is a Critical caught in review that would have shipped the feature dead on the site it was written for.

## The Critical

`Canonical::currentUrl()` built the current URL as `home_url( '/' . $wp->request )`. Under WPML's directory-per-language mode `$wp->request` still carries the language directory (`cs/blog/page/2`), and WPML filters `home_url()` to prepend that directory to whatever path it is handed — it cannot know one is already there. The result was `/cs/cs/blog/page/2/`, which never matches the plugin's canonical, so the guard skipped pagination on every paginated request.

It failed safe — no wrong tag — but it silently disabled the feature.

The motivating site runs WPML in directory mode with five languages: English at the root, `/cs/`, `/pl/`, `/sk/`, `/it/` prefixed. So this would have left four of its five languages unfixed. The fifth is English, at the root with no prefix — and the broken canonical measured at the start of this work was on an English route, so verifying the fix there would have shown a success.

Fixed by not asking `home_url()` for a path at all: the host comes from `home_url()` with no argument (still language-correct under domain-per-language WPML, and with no path there is nothing to double), and the path comes from the request. No WPML detection, no plugin schema this package does not own.

**The tests are the substance of that commit, not the three lines of code.** This shipped because the `home_url()` stub modelled no prefixing. The new stub prepends a language directory the way WPML's filter does, plus round trips for both directory and domain mode. Restoring the old implementation against the new suite fails 6 of 12 tests.

## The two fixes that prompted it

- `Pagination::append()` hardcoded `/page/`. WordPress keeps that in `$wp_rewrite->pagination_base` and localisation plugins change it; a site serving `/blog/strana/2/` would have received a canonical pointing at a 404 — worse than the defect being fixed, because a wrong canonical is an active instruction. The base is now a parameter with a `'page'` default, supplied by a WordPress-aware `PaginationBase::current()`, so `Pagination` stays pure.
- A manual canonical is no longer paginated. An editor pointing `/blog/` at `/campaign/` means every page of the listing to point there; appending `/page/2/` produced a URL that usually does not exist, and the unfiltered behaviour was the correct one. Solved without reading either plugin's storage: append only when the canonical, pagination stripped, still describes the current request.

## The default flip

`$seo_canonical_pagination` now defaults `true`, at the owner's explicit request and with his approval recorded. This is a deliberate exception to `AGENTS.md`'s rule that new rendered-output behaviour defaults off, and it is written up as an exception in `AGENTS.md`, the flag docblock, `README.md` and `CHANGELOG.md` rather than done quietly. The test pinning the default was inverted, not deleted.

Safe because on a site whose listings are real post-type archives the filter is a no-op: `append()` is idempotent and re-derives the identical string.

## Known, deferred

- `Pagination::append()` strips a trailing `/page/N/` unconditionally, so a manual canonical genuinely ending in `/page/5/` is rewritten even on a page-1 request. Predates these commits, narrow, own ticket.
- Yoast is detected by two symbol sets: `Plugin::active()` uses `WPSEO_VERSION`/`WPSEO_Options`, `WarmupSitemap::isYoastSitemapActive()` keeps `WPSEO_VERSION`/`WPSEO_Sitemaps`. Deliberate restraint when consolidating; worth unifying later.

CI green on all six checks, PHP 8.3 and 8.4.

parisek:
## Live verification on a real site

Ran against a DDEV instance of the motivating site: WordPress with Yoast 28.3, WPML, `wp-seo-multilingual`, five languages. The kit's `src/` was rsynced over the vendored copy, so this is the branch's actual code, and the flag was left at its new default rather than set by hand — so the default is part of what was tested.

**It did not work on the first run.** Every route still emitted the unpaginated canonical.

Cause, measured with probes on `wpseo_canonical` rather than guessed:

```
priority  9 (before everything):  https://sloneek.ddev.site/cs/blog/
priority 11 (after everything):   https%3A%2F%2Fsloneek.ddev.site%2Fcs%2Fblog%2F   parse_url host = none

callbacks at 10, in order:
  WPML_WPSEO_Filters::canonical_filter      <- percent-encodes the whole URL
  Parisek\TimberKit\Seo\Canonical::filter   <- ours, second
```

`wp-seo-multilingual` percent-encodes the canonical. `parse_url()` finds no host in that, so the guard read it as a mismatch and passed the value through. It failed safe — no wrong tag was ever emitted — but the feature was a no-op on every WPML + Yoast install, which is most of the fleet this package serves.

This is the **second** time the manual-canonical guard silently disabled the feature it protects; the first was the WPML directory-prefix bug fixed in `f5f5495`. Both passed the full unit suite. Both were only visible on a real site. That is worth recording as the lesson of this branch: the guard's inputs come from other plugins, and stubs modelled what those plugins ought to send rather than what they do.

Fixed in `c86bc6e` — one `urldecode()` retry when the first parse yields no host, keyed on the shape of the value, not on detecting WPML or depending on filter priority.

## Measured after the fix

Canonical on each route, read from the rendered HTML:

```
/blog/                    https://sloneek.ddev.site/blog/            (page 1 untouched)
/blog/page/2/             https://sloneek.ddev.site/blog/page/2/
/blog/page/3/             https://sloneek.ddev.site/blog/page/3/
/cs/blog/page/2/          https://sloneek.ddev.site/cs/blog/page/2/
/pl/blog/page/2/          https://sloneek.ddev.site/pl/blog/page/2/
/sk/blog/page/2/          https://sloneek.ddev.site/sk/blog/page/2/
/it/blog/page/2/          https://sloneek.ddev.site/it/blog/page/2/
/case-studies/page/2/     https://sloneek.ddev.site/case-studies/page/2/
```

Non-listing routes unchanged: `/`, `/ebooks/`, `/cs/o-nas/` all still self-reference their own URL.

### All three WPML negotiation modes, each measured

| Mode | How | Result |
| --- | --- | --- |
| Directory (`/cs/`) | the site's real configuration | correct on all five languages |
| Parameter (`?lang=cs`) | negotiation type switched to 3 | `/blog/page/2/?lang=cs` — query string preserved, segment appended |
| Domain per language | type switched to 2, a second DDEV hostname added | `sloneek-pl.ddev.site/blog/page/2/` self-references its own domain, not the default one |

### The guard still guards

A manual canonical was set on the Blog page (`_yoast_wpseo_canonical` = `/campaign/`), then removed:

```
with manual canonical:    /blog/  and  /blog/page/2/  both -> https://sloneek.ddev.site/campaign/
after removing it:        /blog/page/2/               -> https://sloneek.ddev.site/blog/page/2/
```

No `/campaign/page/2/` was produced. The guard was made robust, not traded away.

The instance was restored afterwards: WPML settings, `.ddev/config.yaml`, and the vendored package back to `v1.43.0` — confirmed by the canonical returning to its broken value.

CI green on all six checks.

parisek:
## Live verification under All in One SEO

The earlier live run covered Yoast only. AIOSEO is the migration target this package was built for, so it was the larger untested surface. Measured on a second project's DDEV instance: AIOSEO Pro + WPML directory mode, default language `cs`, running timber-kit **v1.28.0**.

Deliberately not the whole branch: only `src/Seo/` was dropped into the vendored v1.28.0 and the filter registered from a temporary mu-plugin. Overwriting fifteen minor versions of `StarterBase` would have tested a chimera. This tests the Seo namespace against a real AIOSEO, which is the part that had never run.

First result, before measuring anything else:

```
Plugin::active() = 'aioseo'    has_filter('aioseo_canonical_url') = true
```

The adapter resolves and hooks. That is the first time it has done so outside a stub.

### With and without the filter, same URLs

```
                          without our filter                    with our filter
/cs/clanky/page/2/        …/cs/clanky/            (broken)      …/cs/clanky/page/2/   correct
/cs/clanky/               …/cs/clanky/                          …/cs/clanky/          page 1 untouched
```

So the defect reproduces under AIOSEO exactly as it does under Yoast, and the fix corrects it. The adapter is no longer theory.

### A pre-existing AIOSEO defect this did NOT touch, and should not have

```
/cs/page/2/    canonical = https://mairateam.ddev.site/cs/2/     with AND without our filter
```

AIOSEO emits `/cs/2/` for the front-page pagination — the `page` segment is dropped, producing a canonical to a URL that does not exist. It does that on its own; the value is identical with our filter disabled, so this is not something this branch introduced.

Our guard refused to act on it, which is the designed behaviour and the right one here: the canonical it received (`/cs/2/`) does not describe the request (`/cs/page/2/`), so it passed through untouched rather than compounding a value it cannot explain. A guard that "helpfully" rewrote it would have been guessing at another plugin's output.

Worth its own investigation on that project — it is a live SEO defect there — but it is not this PR's, and this PR correctly leaves it alone.

### Coverage after both live runs

| Surface | Status |
| --- | --- |
| Yoast, WPML directory, five languages | measured |
| Yoast, WPML parameter (`?lang=`) | measured |
| Yoast, WPML domain-per-language | measured |
| Yoast, manual canonical guard | measured |
| **AIOSEO, WPML directory** | **measured** |
| AIOSEO, WPML parameter / domain | not measured — the negotiation mode is handled above the adapter, in `Canonical`, and is plugin-independent by construction |
| No SEO plugin | unit-tested only (`Plugin::active()` returns null, `match` no-ops) |

Both instances were restored: the mu-plugin and the copied `src/Seo/` removed, and the canonical confirmed back to its broken value afterwards.

Issue #112, opened 2026-08-11
README covers a subset of the package — two CLI commands and four classes are undocumented


parisek:
The README documents the package as it stood several releases ago. Features have landed since, each with a PR body and a changelog entry, and none of them reached the one file a consumer actually reads first.

This is not a cosmetic gap. A theme author who cannot see a feature in the README will write it by hand in the theme — which is how the same helper ends up forked across five sites.

## What is missing

**CLI commands.** Two of the five registered commands appear nowhere:

| Command | In README |
| --- | --- |
| `wp timber-kit prune-originals` | yes |
| `wp timber-kit updates` | yes |
| `wp timber-kit convert-utf8mb4` | yes |
| `wp timber-kit acfml-sync-preferences` | yes |
| `wp timber-kit wpml-cleanup-theme-domain` | **no** |
| `wp timber-kit outage-screen` | **no** (new — parisek/timber-kit#111) |

**Classes with no section at all:**

- `BreezeWarmupSitemap` — cache warm-up after a purge
- `VideoCodecs` — the codec-string derivation behind `formatVideo()`
- `OutageDropIns` — the generator behind the new command
- `Updates\*` — `wp timber-kit updates` is mentioned, but the update-file anatomy that a consumer has to write is not

**Mentioned once, in passing, with no usage:** `MenuData`, `OriginalImagePruner`.

## Suggested shape

Rather than one long append, the README needs a decision it does not currently make: what belongs in it, and what belongs in `docs/`.

Proposal — the README answers *what exists and when I would reach for it*, one short section per feature with a single worked call, and links to `docs/adr/` for the reasoning. Anything that needs more than ~20 lines to explain (the update-file anatomy, the breadcrumb's options-page contract) gets a `docs/` page and a one-line pointer.

A `## Command-line` table listing all five commands in one place would also stop this drifting again: a new command that is not in the table is visibly missing, where a missing prose section is not.

## Not in scope

Rewriting the sections that are already accurate. This is about coverage, not style.

PR #162, opened 2026-08-29
feat(health): check the resizer can write its output format, with alpha

parisek:
## From which project

`sloneek` on Cloudways. The site's transparent hero logo rendered on a black box; the cause was ImageMagick writing AVIF without an alpha channel. It came back a second time after a host upgrade landed ImageMagick 7.1.2-31 **without** the AVIF delegate — `queryFormats()` returned 260 formats with none of the HEIF family, and `setImageFormat("avif")` threw `Unable to set image format`.

Neither state was visible anywhere in wp-admin. Both were found by hand, months apart.

## Why

`timber_kit_resizer_formats` already asks which formats the backend can **decode**, so uploads get cropped instead of served full-size. Nothing asked about the **output** side — and the output format is the one every image on the site passes through. One missing delegate takes down every variant, not one editor's upload.

The failure is silent by construction. `Resizer::resizeImage()` catches the encoder throw, `error_log()`s it and returns `null`, so the variant simply leaves `<picture>`. No fallback, no warning. On `sloneek` the only reason nothing broke visibly was a cache of 2628 files copied from the previous host.

## What it does

A new `METHOD_EFFECT` check, `resizer_output_format_writable`, encodes a half-transparent 16x16 image in the configured output format and reads a pixel back.

**A capability list cannot replace the encode.** ImageMagick 6.9.11 listed AVIF and wrote AVIF — it just flattened alpha. That is the failure this started from, and only a round trip sees it.

Four verdicts stay distinct, because they send an admin to four different fixes:

| Verdict | Status | Means |
| --- | --- | --- |
| `no_backend` | critical | Neither Imagick nor GD — nothing resizes at all |
| `missing_delegate` | critical | Format absent from the build; rebuild with libaom/libavif/libheif |
| `write_failed` | critical | Listed but refuses to encode — decoder present, encoder missing |
| `alpha_lost` | critical | Encodes, but transparency comes back opaque — outdated encoder |
| `unverified` | recommended | Encoded, but the backend cannot read its own output back, so transparency is unchecked |

Details that matter:

- **Backend selection mirrors production exactly** — `class_exists(Imagick::class) ? Imagick : GD`, the same rule `Spatie\Image` and `Resizer::probeBackendFormats()` use. A probe hardcoded to Imagick would report failures a GD-only host never hits, and a false critical is worse than no check.
- **The format comes from `timber_kit_resizer_target_format`**, not the constant behind it, so a project that switched to WebP is told about WebP. It is probed **verbatim**: `Resizer::normalizeFormat()` lowercases only per-variant formats, so a filtered `AVIF` reaches the encoder uppercase, and probing a tidied copy would test a format production never uses.
- **"No backend at all" is asked before the format question**, so a project that filtered the output to PNG cannot get a green row on a host where nothing resizes.
- **Only the request-wide format is probed.** A per-variant `'format' => …` override is not; a health check cannot enumerate every format a template might ask for.
- **JPEG/PNG/GIF short-circuit** without an encode — no delegate can be missing for them.
- **Nothing is cached.** See rejected alternatives.

## Rejected

- **A transient keyed by the ImageMagick version.** Measured on the affected host: 5.87 ms per probe, in memory, no temp file. Site Health direct tests run on one admin screen and in the weekly cron, so the cache would buy milliseconds and owe a staleness bug.
- **Falling back to a writable format when the probe fails.** That changes what `<picture>` emits and belongs in `Resizer` with its own decision. This PR only makes the failure visible.
- **One shared failure message.** "Lists the format but drops alpha" reads as a working server until somebody opens an image; collapsing it into "AVIF unavailable" sends the admin to the wrong fix.
- **`protected function probe()` as the test seam.** Checks in this registry are `final`, so the seam is an injected `ImageFormatProbe` — which also lets a future CLI adapter reuse the verdict without wp-admin, as `SiteHealthAdapter`'s docblock anticipates.
- **`'image' . $format` for the GD writer.** The format arrives from a public filter; concatenation would let any value select an arbitrary `image*()` function. It is a `match` on literal calls.

## Review

Two independent agent reviewers plus CI; the full record is in a comment on this PR. Both reviewers converged on two real defects (a green row on a backend-less host, and conflating a failed read-back with a failed encode). CI then found a third that neither named: GitHub's runners report `gd_info()['AVIF Support']` as true while `imageavif()` fails, because libavif ships a decoder and no encoder — so the check was right about that host and the test guard was wrong.

## Tests

23 new tests. The stubbed set asserts the judgement; `BackendImageFormatProbeTest` runs the **real encoder**, which is where the risk is.

Included deliberately: a **negative control** that encodes a genuinely opaque image and asserts it is caught as `alpha_lost`. Without it every other assertion would still pass with the threshold inverted, because a working backend never produces the failing case.

Per-format cases skip when the running build cannot write that format — decided by **attempting an encode**, never by reading `gd_info()`. Predicting an encode from a capability flag is the exact mistake this check exists to catch, and the first version of the test made it.

## Verification

`composer check` green — 1909 unit tests, 20 property tests, PHPStan level 8, ADR index. The one pre-existing warning and skip are present on `main` too.

PHPStan caught two real defects during the work: `getColor(true)` where the signature is `0|1|2`, and the variable-function invocation noted above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -434,6 +434,36 @@ mid-process must call `Helpers::flushFieldGroups()` itself.
 A screen `wp_json_encode()` cannot encode is never memoized, because casting
 `false` to a string would collapse every such screen onto one key.
 
+### Fixed
+
+- `StarterBase::cleanup_cached_images()` deleted resizer derivatives that other
+  attachments were still using. The cache under `wp-content/cache/image` is
+  keyed by file, but one file routinely carries several attachment rows: WPML
+  writes one per language, and a duplicate upload can be pointed at a path that
+  already exists. Deleting any one of those rows wiped the shared derivatives,
+  and the site then regenerated them — or, where the encoder was broken, served
+  a damaged image.
+
+  Measured on a five-language site: 5542 files were shared by 25981 attachment
+  rows, so roughly four fifths of the media library could take a live image
+  down with it. The homepage hero lost its derivatives while five attachment
+  rows and the rendered page still referenced them.
+
+  The delete now runs only when no other attachment points at the same
+  `_wp_attached_file`. It fails closed: where that question cannot be answered
+  the files are kept, because a stale derivative is overwritten by the next
+  resize while one deleted in error vanishes from a page still serving it.
+  "Cannot be answered" covers a missing `$wpdb`, an empty `_wp_attached_file`
+  (a filter supplied the path, so siblings have no key to match on), and a
+  failed query — `get_var()` reports an error by returning null, which casts to
+  the same zero a genuine "no siblings" answer gives, so the null and
+  `last_error` are both checked rather than read as a count.
+
+  Not fixed here, and tracked separately: two different files that share a
+  basename across upload-year folders still collide in the flat cache
+  namespace, so deleting one can remove the other's derivative. That needs a
+  cache-naming change, not a guard.
+
 ## [1.42.0] - 2026-08-26
 
   310 calls covered **33 distinct URLs** — the same menu and options-page links
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   told about WebP. Delegate-free formats (JPEG/PNG/GIF) short-circuit without
   an encode.
 
+  The verdict is measured on a real build, so a host that reports AVIF support
+  and cannot encode it is caught. GitHub's runners are exactly that host:
+  `gd_info()['AVIF Support']` is true there while `imageavif()` fails, because
+  libavif ships with a decoder and no encoder.
+
   Encoding and decoding are judged separately. A build that writes the format
   but cannot read its own output back reports `unverified` (recommended), not a
   failure — the resizer still produces valid files there, and calling that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+
+- Site Health — `resizer_output_format_writable`. The existing
+  `timber_kit_resizer_formats` test asks which formats the backend can
+  *decode*; nothing asked whether it can **write** the format the resizer
+  targets, and that is the one every image passes through. When the encode
+  fails, `Resizer::resizeImage()` logs and returns null, so the variant drops
+  out of `<picture>` with no fallback and nothing visible in wp-admin.
+
+  The check encodes a half-transparent 16x16 image in the configured format and
+  reads a pixel back, because a capability list cannot see the failure that
+  prompted it: Cloudways' ImageMagick 6.9.11 listed AVIF and wrote AVIF, and
+  flattened the alpha channel while doing so. Four verdicts stay distinct — no
+  backend, missing delegate, read-only delegate, dropped alpha — because they
+  send an admin to four different fixes.
+
+  Backend selection mirrors `Spatie\Image` and `Resizer::probeBackendFormats()`
+  exactly (Imagick when the class exists, GD otherwise), so the probe cannot
+  report a failure production never hits. The format is read through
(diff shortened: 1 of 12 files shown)


parisek:
## Agent review — 2026-08-29

Two reviewers, independent, different lenses and different models: **Codex** (correctness: alpha semantics, resource leaks, false verdicts) and **Claude Opus 5** via the `code-reviewer` agent (integration and project doctrine). Neither saw the other's output. GitHub Actions was effectively a third reviewer and found the defect both had reasoned about.

### Where they converged

Both flagged the same two defects, from opposite directions. That agreement is the strongest evidence here, and both were real:

1. **The delegate-free short-circuit returned `good` on a host with no image backend.** `VERDICT_NO_BACKEND` was unreachable for any project that filtered the output format to PNG or JPEG — on a host where nothing resizes at all. Fixed: `hasBackend()` is asked before the format question.
2. **Verifying the encode by reading the file back conflated two capabilities.** A build that writes a format it cannot read reported `write_failed`. Fixed: `VERDICT_UNVERIFIED` at `recommended`.

CI then proved #2 was not hypothetical — though not for the reason either reviewer gave. See below.

### Codex — correctness

Cleared: Imagick `getColor(1)` alpha semantics, GD's `>> 24 & 0x7F` packing, the sample point at (12,8), and resource cleanup. I had specifically asked it to verify the alpha direction against real library behaviour rather than assume, since an inverted threshold would report healthy servers as broken.

Fixed beyond the convergent two:

- **`imagecreatetruecolor()` unguarded.** Under memory pressure it returns `false`; the first typed GD call throws a `TypeError`, and `imagedestroy( false )` in the `finally` throws a second one on the way out. That escapes `probe()` and fatals the Site Health screen — a health check taking down the page that reports health. The guard existed in an earlier draft and was lost when the probe moved to its own class.
- **Case handling did not match production.** `Resizer::normalizeFormat()` lowercases only a *per-variant* format; the request-wide one from the filter reaches the encoder verbatim (`src/Resizer.php:184`, `:699`). The check lowercased before probing, so a filtered `AVIF` was tested as `avif` and passed while Spatie's GD driver rejects the uppercase spelling at render time. Now probed exactly as written.

Deferred, with reasoning: **per-variant `'format' => …` overrides are not probed.** A health check cannot enumerate every format a template might request. The request-wide format is what every image uses by default, and the scope is now stated in the class docblock.

### Claude Opus 5 — integration and doctrine

- **README was not updated.** `AGENTS.md` makes this a hard rule for a PR adding a public class, and cites four classes that went missing this way (#112). I had missed it. Added, following the `preload_chain_healthy` section's shape.
- **Namespace.** `Health\` top level holds only framework (`HealthCheck`, `Result`, `CheckRegistry`, `SiteHealthAdapter`); check collaborators live in a subnamespace, as `Utf8mb4Tables` does with `Health\Db\`. Moved to `Health\Image\` — cheap now, a breaking change after a release.
- **`public static alphaVerdict()` was API existing only for a test.** Now `private`, behind a named `ALPHA_OPAQUE_THRESHOLD`. The negative control moved to the public `probe()` using `jpeg`, which cannot carry alpha — and that immediately caught a real bug the private seam had hidden: a writer added to the format map but not to the dispatch `match`. The two lists have since been collapsed into one.

Rejected: **gating the check on `$resizer_format_health`.** That flag's own docblock scopes it to which formats the backend can *decode*; encode is a different axis, and a project silencing upload-format noise should not thereby silence a site-wide critical. Took the reviewer's alternative instead — the label now names the axis ("Image resizer output format is writable").

### What CI caught that neither reviewer did

The first push was red, and stayed red after the encode/read-back split. That ruled out the read-back: **the encode itself fails on GitHub's runners.** `gd_info()['AVIF Support']` is `true` there while `imageavif()` fails, because libavif ships with a decoder and no encoder.

So `write_failed` was the *correct* verdict on that host, and the defect was in my test guard — it predicted an encode from a capability flag. That is precisely the mistake this check exists to catch, made inside the test that verifies it. The guard now attempts a real encode and skips when that fails.

Worth stating plainly: the reviewers reasoned their way to "encode and decode are separate capabilities" and were right about the principle, while the concrete failure in front of us was a third case neither had named.

PR #167, opened 2026-09-08
fix(starter-base): actually remove global styles, behind an opt-in flag

parisek:
## From which project

`rezidence-ponavia`, phase F of its StarterBase migration. Reviewing `Base.php` method by method turned up an override of this method — and then the discovery that neither the override nor the parent worked.

## The bug

`remove_global_styles_and_svg_filters()` has never removed anything.

WordPress registers the callback **twice**. From `wp-includes/default-filters.php`:

```php
add_action( 'wp_enqueue_scripts', 'wp_enqueue_global_styles' );
add_action( 'wp_footer', 'wp_enqueue_global_styles', 1 );
```

The method removed only the `wp_footer` one, and **without a priority**. `remove_action()` matches on the (hook, callback, priority) triple, so the default `10` silently misses a callback registered at `1`.

Proven with WordPress's own hook API on a throwaway hook, not a reimplementation:

```
registered at priority 1:            YES
remove_action WITHOUT priority:      NO    <- returns false
still registered afterwards:         YES (priority 1)
remove_action WITH priority 1:       YES
still registered afterwards:         no
```

So every theme calling this shipped `global-styles-inline-css` in the page while the method name said otherwise. Fixed: both registrations removed, priority passed.

## Why the fix is opt-in

Making it work **is** a behaviour change, and a bigger one than it looks.

The stylesheet is ~10 kB — `--wp--preset--*` custom properties plus the `.has-*-color` / `.has-*-font-size` utility classes Gutenberg writes into post content. A Tailwind theme has no use for WP's stock palette, gradients, shadows or spacing scale. **But the same stylesheet is what makes an editor's colour and font-size picks render.**

Measured on `rezidence-ponavia`, on a published page, in a browser:

| | `<h6 class="has-small-font-size">` |
| --- | --- |
| with the stylesheet | **13px** |
| without it | **24px** |

No error, no warning. Just a heading that changed size on a privacy page nobody would think to re-check.

Then the fleet. Of 105 projects with a committed database dump, **55 carry content this would change**:

```
besteto          8458 occurrences
skialpyjakvino   4992
debra            2497
umbili           1248
ebeton            903
albaform          836
```

A default of `true` would have been a silent regression on more than half the fleet, discovered by clients. So:

```php
protected bool $remove_global_styles = false;
```

Default `false` preserves today's *effective* behaviour exactly — the method simply stops pretending. A project switches it on after checking its own content:

```
wp db query "SELECT COUNT(*) FROM wp_posts \
  WHERE post_content REGEXP 'has-[a-z0-9-]+-(color|font-size)|is-layout-constrained'"
```

Zero means the stylesheet is inert there and this is ~10 kB off every page. `rezidence-ponavia` returns **4**, so it is NOT switching it on — the `<h6>` above has to be rewritten to a Tailwind class first. That is the intended shape of the decision: per project, against that project's content.

## Two variables that are ours, not WP's

Worth knowing before anyone reads the preset list and concludes the whole block is junk. Alongside the stock palette sit:

```
--wp--style--global--content-size: 645px
--wp--style--global--wide-size:   1340px
```

Those come from the theme's own `theme.json` `settings.layout`, and WP's constrained-layout support reads them. On the page I measured, their only consumers are inside the same stylesheet — so removing the block removes both the declarations and their uses, leaving no dangling `var()`. A project using `core/group` with constrained layout should still verify; `is-layout-constrained` is in the check query above for exactly that reason.

## Verification

Both branches, against a live DDEV site with the patched package in `vendor/`:

| `$remove_global_styles` | `global-styles-inline-css` in page |
| --- | --- |
| default (`false`) | present — unchanged from before this PR |
| `true` | absent, page is 10366 bytes smaller |

`php -l` clean. Registration is gated in `registerMediaHooks()` alongside the other media flags, so a consumer on the default never registers the `init` callback at all.

## Downstream note

26 fleet projects run StarterBase today. None of them changes behaviour on this bump, by design. The upgrade note is the check query — projects that come back zero get 10 kB per page for one line of config.

https://claude.ai/code/session_01PT37YMD6axaf1yhU8XoUk9


parisek:
## Narrowed the scope — thanks, that was a fair objection

The first revision of this branch did two things I bundled that should not have been bundled:

**1. It added a removal for `wp_global_styles_render_svg_filters` that the kit never had** — and that removes nothing. Verified on WP 7.0.4:

```
wp_body_open   has_action() -> false
wp_footer      has_action() -> false
wp_head        has_action() -> false
function_exists() -> true
```

The callback is registered on no hook at all. It was deprecated in WP 6.3 and is handled per-block now, exactly as the method's own comment said. So the line was dead code.

**2. It put that removal behind `$remove_global_styles`.** Even if it had done something, SVG filters are a different concern from the presets stylesheet, with a different risk profile. The flag's documented risk is "editor-chosen colours and font sizes silently change" — an SVG-filter removal has nothing to do with that, and a flag that means two things is a flag nobody can reason about.

Both dropped. What remains is one concern: the removal never worked, make it work, gate it because working is a behaviour change.

The two removals that stay are each necessary — both registrations exist, verified live rather than read off `default-filters.php`:

```
wp_enqueue_scripts   registered @10  -> removal needed
wp_footer            registered @1   -> removal needed
```

The method keeps its `_and_svg_filters` name with a comment saying why it does nothing about them. Renaming a public method is a breaking change and does not belong in a fix.

Re-verified after narrowing, against a live site with the patched package in `vendor/`: default leaves `global-styles-inline-css` in place; `$remove_global_styles = true` removes it, stable across three consecutive requests (the first check read `2` and was opcache, not behaviour — worth knowing when testing this by hand).

PR #171, opened 2026-09-09
fix(block-renderer): stop caching blocks whose render expanded a shortcode

parisek:
## Problem

`BlockRenderer` refuses to cache a block with side effects, but it detects them only by comparing the script and style queues around the render:

```php
$has_side_effects = ( array_diff( wp_scripts()->queue, $scripts_before ) || … );
```

That misses WPForms entirely. `WPForms_Frontend::output()` enqueues nothing — it only appends the form to its own `$forms` array (`src/Frontend/Frontend.php:350`). The enqueue happens later, on `wp_footer` priority 15:

```php
public function assets_footer(): void {
    if ( empty( $this->forms ) && ! $this->assets_global() ) {
        return;
    }
```

So a block whose ACF `post_object` resolves a WPForms form leaves both queues untouched, passes the guard, gets stored, and from the next request on the render callback never runs. The cached HTML still contains the form markup, so the page looks correct — and the form does nothing.

`writeToCache()`'s own docblock already stated the intent: *"form-plugin enqueues happened during compile, caching would skip the enqueues on next request and break the form."* The intent was right; the detection was incomplete.

## Evidence

Measured on a live site. WPForms 2.0.1.1, timber-kit 1.46.0, Redis with `flush_group`, `global-assets` off:

```
wp cache flush-group acf_block_91
load 1: wpforms-js=12  form-markup=1
load 2: wpforms-js=0   form-markup=1
load 3: wpforms-js=0   form-markup=1
```

A sibling project on the same server, with the same kit and plugin versions, is unaffected — but only by accident. Its form block happens to register a `block_<slug>_content` filter, and `$has_dynamic_filter` disables the cache for any block that has one. Remove that filter and it breaks identically, with no warning.

Probe of one real block render, with the ACF meta context set the way ACF sets it:

```
delta=1               // Helpers::dynamicFormatCount() rose
form_html_len=6993    // the markup that went into the cache
wpforms_registered=1  // the side effect the queue comparison cannot see
```

## Fix

Compare `Helpers::dynamicFormatCount()` around the render as a second, independent test.

That counter already exists, `MenuFieldsCache` already gates its own writes on it, and its docblock states the principle this needs: *"Every increment is decided from an INPUT, never from whether an output differed. Observing the output cannot prove purity."*

It rises whenever a field expanded a registered shortcode. So it covers a `post_object` resolving WPForms or Contact Form 7, and equally covers a shortcode an editor typed into a WYSIWYG field — a case no field-name check reaches.

**Cost:** a block holding any registered shortcode is no longer cached, harmless ones included. That is the safe direction, and skipping a cache write is cheaper than serving a dead form. `timber_kit/block_renderer/use_cache` remains the per-block override.

**Scope:** only projects whose production runs an external object cache with `flush_group` support. Without one the cache path was never active.

## The existing test asserted nothing

`test_side_effecting_block_excluded_from_cache` injected its output through the `empty_alert_html` filter. That sets `$rendered_empty_alert`, and `writeToCache()` skips the write on that flag alone:

```php
if ( $use_cache && ! $has_side_effects && ! $rendered_empty_alert ) {
```

So `wp_cache_set` was never called, with or without the guard. My first attempt at the new test made exactly the same mistake and passed against unpatched code, which is how this surfaced.

Both tests now compile a real Twig template so `$template_output` is genuine block output, and both assert on what was written rather than on a `never()` expectation. Each was mutation-checked: removing its own guard makes it fail, and it alone.

## Verification

- `composer check` green — 1974 unit tests, 20 property tests, PHPStan clean, ADR index OK.
- Warning and skip counts unchanged from `main` (1973 → 1974 tests, 1 warning, 1 skipped before and after).
- New test against unpatched `src/`: fails, showing the poisoned entry.

```
-Array &0 []
+Array &0 [
+    'acf_block_531302…' => '<div><div class="wpforms-container">[wpforms id="200"]</div></div>',
+]
```

## Rejected

- **Reading `wpforms()->frontend->forms` in `BlockRenderer`.** Names one plugin in a class that should not know about any, and covers less: a shortcode in a WYSIWYG field hits the same bug and no plugin-specific check sees it.
- **Keeping the cache and re-arming WPForms on a hit.** `assets_footer()` also passes `$forms` to the `wpforms_wp_footer` action, which addons read for per-form data. Forcing the assets with an empty array is a second, quieter wrong state.
- **Fixing it per project via `timber_kit/block_renderer/use_cache`.** Works, and is deployed on the affected site as a stopgap, but every downstream has to discover the bug first. A fleet scan found 21 blocks across 17 projects with the same shape and no `_content` filter.

https://claude.ai/code/session_01JvmJ9ytv8WUA34Dr8DhMoe

---

## Update after review (round 1)

Reviewed by Fable (correctness/downstream lens) and Codex (adversarial-verification lens), 2026-09-09, run independently. Round 1 found two false claims and one live gap in the first commit. All three are fixed in `b5f3f41`.

**Fixed — WYSIWYG coverage was claimed and did not exist.** `formatFields()` calls `get_field_objects()` with ACF's default `$format_value = true`. ACF's WYSIWYG `format_value()` applies `acf_the_content` (`class-acf-field-wysiwyg.php:410`), which carries `do_shortcode` at priority 11 (`:79`). A `[wpforms id="…"]` typed into a WYSIWYG field is already expanded before any value reaches `expandShortcodes()`, so no bracket is left, the counter stays flat, and the block was still cached. Verified against the ACF Pro source. The guard now also watches core's own `do_shortcode_tag` filter (`wp-includes/shortcodes.php:447`), which fires for every shortcode WordPress actually executes — covering ACF's expansion, Twig-side `do_shortcode()` and `field_formatter_<type>` callbacks, without naming a plugin.

**Fixed — `use_cache` was offered as an escape hatch and cannot be one.** `writeToCache()` writes only when `$use_cache && ! $has_side_effects && ! $rendered_empty_alert`, so forcing `use_cache` on leaves a rejected block rejected. Added `timber_kit/block_renderer/has_side_effects`, which actually overrides the verdict, with a test.

**Fixed — blast radius was wrong in both directions at once.** With the WYSIWYG hole open the real cost was near nil while the stated coverage was too broad. Now the cost is real (a block holding `[caption]` or `[embed]` stops being cached) and the coverage matches the code. Restated in the CHANGELOG and the docblock.

**Dropped — an unverified Contact Form 7 claim.** CF7 is not installed anywhere I could measure, and it enqueues on `wp_enqueue_scripts` rather than deferring to `wp_footer`, so the sentence was probably wrong as well as unsupported.

**Not changed — latent holes with no consumer.** `content_data` replacing `formatFields()` entirely, and `post_object` with `return_format: id` or a `relationship` field pointing at a form post. A grep of all themes found no project registering `content_data` or a real `field_formatter_*` callback. Noted rather than coded around.

**Reviewer error, corrected.** Round 1 reported `RenderTest.php` failing intermittently (2 of 3 runs). That was my setup mistake, not a flaky test: both reviewers were pointed at the same working tree and one was mutating `src/BlockRenderer.php` while the other ran the suite. On a clean tree the file passes 20/20.

### Guards and their tests

Three independent tests, each mutation-checked — removing one guard fails exactly one test, and a different one each time:

| Guard | Test |
| --- | --- |
| script/style queues grew | `test_side_effecting_block_excluded_from_cache` |
| `dynamicFormatCount()` moved | `test_wpforms_block_is_not_cached_although_the_queues_never_move` |
| `do_shortcode_tag` fired | `test_block_is_not_cached_when_acf_expanded_a_shortcode_before_helpers_saw_it` |

Plus `test_has_side_effects_filter_restores_caching` for the new escape hatch.

`composer check` green: 1976 unit tests, 20 property tests, PHPStan clean, ADR index in sync. Warning and skip counts unchanged from `main`.


parisek:
## Agent review record

Two reviewers per round, run independently in separate checkouts so neither could see the other's output: **Fable** (correctness and downstream-impact lens) and **Codex** (adversarial-verification lens), 2026-09-09.

### Round 1 — on `78a9685`

| Finding | Reviewer | Disposition |
| --- | --- | --- |
| WYSIWYG coverage was claimed and did not exist — ACF expands shortcodes inside `get_field_objects()` via `acf_the_content` + `do_shortcode` priority 11, so the counter never moves | Fable | **Fixed** in `b5f3f41`. Verified against ACF Pro source before acting |
| `use_cache` documented as the escape hatch, but `writeToCache()` needs `$use_cache && ! $has_side_effects` | Fable | **Fixed** — new `timber_kit/block_renderer/has_side_effects` filter |
| Blast radius overstated and coverage overstated at the same time | Fable | **Fixed** in CHANGELOG and docblock |
| Unverified Contact Form 7 claim | Fable | **Dropped**. CF7 is not installed anywhere measurable here, and it enqueues on `wp_enqueue_scripts` rather than deferring |
| Latent holes with no consumer: `content_data` replacing `formatFields()`, `post_object` with `return_format: id`, `relationship` fields | Fable | **Noted, not coded around.** A grep of all sibling themes found no project registering `content_data` or a real `field_formatter_*` callback |
| Counter moves on the WPForms/CF7 `post_object` path; both guards mutation-sensitive | Codex | Confirmed the author's claims |

Codex's round-1 run ended without a written report — only a command log, truncated mid-work. Treated as a partial result, not a second independent voice, and re-run in round 2 with an explicit report requirement.

### Round 2 — on `b5f3f41`

**Converged (both reviewers, independently):**

| Finding | Disposition |
| --- | --- |
| No `try`/`finally` — a throwing template leaks the probe onto the global hook for the rest of the request | **Fixed** in `4ef4490` |
| `pre_do_shortcode_tag` short-circuits bypass the probe (`shortcodes.php:427` returns before `:447`) | **Fixed** — the probe moved to the earlier hook, which fires unconditionally. This also makes the "every shortcode" wording true |

**Single-reviewer findings:**

| Finding | Reviewer | Disposition |
| --- | --- | --- |
| `remove_filter` had no test — deleting the call passed 17/17 | Fable | **Fixed.** Reproduced independently before acting; the test now asserts the same closure instance and priority were unhooked |
| Blast-radius wording still too narrow: the probe covers the whole window, so a listing block rendering another post's `[caption]` also stops being cached | Fable | **Fixed** in CHANGELOG and docblock |
| The live-site measurement cannot be reproduced from this repository | Codex | **Accepted.** Labelled as external evidence in the CHANGELOG |

**Disagreement, resolved by measurement:** Codex called the WYSIWYG test a medium defect for having its ACF stub invoke the captured closure rather than running real shortcode dispatch; Fable called it a faithful simulation, not a tautology. Six mutations settle it — wrong hook name, wrong priority, no-op probe body, missing unhook, probe disabled, registration moved after `buildContent()` — each fails that test and only that test. Kept, with Codex's point acknowledged: what it takes on faith is that WordPress calls a registered filter, which is the API contract.

**Reviewer report corrected:** round 1 reported `RenderTest.php` failing 2 of 3 runs. That was the author's setup error — both reviewers were pointed at one working tree and one was mutating `src/BlockRenderer.php` while the other ran the suite. On a clean tree the file passes 20/20. Round 2 gave each reviewer its own clone.

**Not a code defect:** Codex's PHPStan step failed on `tcp://127.0.0.1:0`, a sandbox networking limit. PHPStan is clean locally and in CI.

### Current state

Seven mutations, each failing exactly one test and a different one each time: unhook removed, unhook priority wrong, probe body no-op, probe on the later hook, counter disabled, probe disabled, queue guard disabled.

`composer check`: 1976 unit tests, 20 property tests, PHPStan clean, ADR index in sync. Warning and skip counts unchanged from `main`.

PR #174, opened 2026-09-09
fix(starter-base): reset the query through set_404() rather than by hand

### LINKED ISSUE #173: disable_search flips query flags by hand instead of calling WP_Query::set_404()
Third and last of the `disable_search()` findings from 2026-09-09. The other two are fixed: #169 sent the missing 404 status, #172 stopped the method mutating secondary queries. This one is hardening rather than a reported failure, and it is the reason the release is being held.

## What it does now

```php
$query->is_search       = false;
$query->query_vars['s'] = false;
$query->query['s']      = false;
$query->is_404          = true;
```

Four flags, flipped by hand. `WP_Query` has an API for exactly this:

```php
public function set_404() {
	$is_feed = $this->is_feed;

	$this->init_query_flags();
	$this->is_404 = true;

	$this->is_feed = $is_feed;

	do_action_ref_array( 'set_404', [ $this ] );
}
```

## Why the hand-written version is weaker

- **`init_query_flags()` is skipped.** Core resets every conditional before setting `is_404`. The manual version clears `is_search` and sets `is_404`, and leaves everything else exactly as it was. A query that was also `is_home`, `is_archive` or `is_post_type_archive` keeps those true **alongside** `is_404`, so a template asking `is_archive()` on a 404 gets a yes.
- **The `set_404` action never fires.** Anything hooked there — core, a plugin, a consumer's own theme — does not run. That is a contract this method silently opts out of.
- **`is_feed` is not preserved.** Core deliberately carries it across the reset. The manual version has nothing to carry it across, so today it happens not to matter, and it would matter the moment `init_query_flags()` were introduced without the same care.

## Suggested change

```php
$query->set( 's', '' );
$query->set_404();
```

`set()` clears the search term through the query's own API rather than by writing three properties, and `set_404()` does the rest correctly.

## Why it is not urgent

No failure has been observed from it. The two fixed defects were both reachable and one was measured on a live site; this one is a latent inconsistency that a future template or plugin would trip over rather than something breaking today.

## Testing note

The unit stub in `tests/bootstrap.php` already has `set_404()`, but it does not model `init_query_flags()` or fire the action. Proving this fix properly means teaching the stub what core actually does, otherwise the test asserts the stub's behaviour rather than the fix's. That is the bulk of the work here, and it is the reason this was kept out of #172 rather than bundled in.

Found by Codex while reviewing #172, which rated it real, pre-existing and separate.

parisek:
Fixes #173. Third and last of the `disable_search()` fixes from 2026-09-09, after #169 (missing 404 status, shipped in v1.47.0) and #172 (guard read the globals instead of the object).

The method flipped four query flags by hand where `WP_Query::set_404()` exists. That skipped three things core does:

- **`init_query_flags()` never ran**, so every other conditional survived alongside `is_404`. A template asking `is_archive()` on a 404 got a yes.
- **The `set_404` action never fired**, so anything hooked to it silently did not run.
- **`is_feed` was not preserved** across a reset, because there was no reset.

## The change

```php
$query->set( 's', '' );
$query->query['s'] = '';
$query->set_404();
```

`set_404()` handles `is_search` as part of the reset, so the manual assignment is gone.

**`$query->query` is still blanked, and that is deliberate.** An earlier revision of this PR dropped it, reasoning that nothing downstream reads the raw request args. That reasoning was wrong: `pre_get_posts` fires **after** this hook, and a known plugin idiom — the Algolia-style search integrations among them — reads `$query->query['s']` there. Core's own `set_404()` does not clear it, but this method always has, and dropping it would have been an observable behaviour change rather than a simplification. Review caught it.

## Tests

The `WP_Query` stub in `tests/bootstrap.php` now models what core actually does, because a test written against the old stub would have asserted the stub's behaviour rather than the fix's — the exact trap the issue describes. `init_query_flags()` resets the conditionals **including `is_feed`**, and `set_404()` saves and restores it around the call, matching core's division of responsibility rather than only its end state.

Pinned by tests: a conditional that was true before the call is false afterwards, the `set_404` action fires, `is_feed` survives `set_404()`, and `is_feed` is reset by a direct `init_query_flags()` call. That last one is the case the two stub designs disagreed about.

1979 tests / property suite unchanged / PHPStan level 8 clean / no new notices. `BlockAuthorEnumerationTest` shares the stub and stays green.

## No feature flag

`AGENTS.md` requires behaviour changes behind a `StarterBase` flag defaulting to off. Waived by the owner for this series on 2026-09-09: these correct defects rather than introduce behaviour, and gating a correction behind an opt-in would keep the incorrect behaviour as the default.

## Known and deliberately not addressed

Clearing `s` and marking the query 404 does not stop WordPress running the posts query that follows, so a nominally-404 request still fetches real posts into `$wp_query->posts`. Found during review of this PR. It is **pre-existing** — the hand-written version had it too — and closing it means adding a real no-results constraint, which changes behaviour more than this series should. Tracked separately.


parisek:
Closing this without merging. The change is correct and the CI is green; the reason is value against risk, not quality.

**What it fixes is not a problem anyone has.** The reported defect in this series was the soft 404, and #169 fixed that — measured on a live site, verified end to end. This PR changes the method's internals so it calls core's API instead of writing four properties. The only concrete consequence a consumer would notice is that `is_archive()` stops answering true on a 404. Nobody reported that, and the 404 pages on the site this was found from render correctly today.

**What it costs is the largest behaviour change of the three.** `init_query_flags()` resets every conditional, in a library used by dozens of sites, none of which asked for it. The review that approved it also noted that full WPML extension coverage could not be verified.

**And it did not leave the method coherent.** Reviewing this PR surfaced #175: the query is marked 404 and still runs unconstrained, populating `$wp_query->posts` on a 404 response. Merging this would have tidied the internals while leaving the functional gap open — a worse resting state than either fixing both or fixing neither.

#173 stays open as a documented internal inconsistency. It is real, and it is worth revisiting when there is a reason beyond tidiness: a template that actually trips over a surviving conditional, or work on #175 that makes it natural to do both together.

The diff is preserved on this PR and can be restored if that day comes.

PR #177, opened 2026-09-10
feat(starter-base): close /author/<nicename>/, the vector the other three guards point at

### LINKED ISSUE #176: Author archives at /author/<nicename>/ stay reachable while the other three enumeration vectors are closed
## What happens

`/author/<user_nicename>/` is reachable on every site this kit ships to. WordPress
registers those rewrite rules by default and nothing here removes them.

The kit already closes three username-enumeration vectors and names them as a set:

| Vector | Handled by | Result |
| --- | --- | --- |
| `/?author=N` | `block_author_enumeration` | 404 |
| `/wp-json/wp/v2/users` | `restrict_rest_users` | 401 |
| `/wp-sitemap-users-1.xml` | `disable_author_sitemap` | 404 |
| `/author/<nicename>/` | — | **reachable** |

The fourth is the destination the first one exists to protect. `block_author_enumeration`'s
docblock says core "canonically redirects `/?author=1` to `/author/{username}/`, which
exposes the login slug" — and then leaves that page open. Blocking the redirect while the
target answers is half a guard.

## Measured on a downstream site

`sloneek`, 9 users with published posts, so 9 reachable URLs:

- `/author/alexandra/` answers **503 with `Retry-After: 3600`** (this theme's router has a
  fail-loud fallback for unrouted query states, which is correct for routes that are coming
  and wrong for these)
- no `noindex` — the robots value is only `max-image-preview:large, max-snippet:-1,
  max-video-preview:-1`
- a self-referencing canonical
- `<title>` carrying the person's real name

So a crawler is told the page is temporarily unavailable, may be indexed, and is
canonical. `Retry-After` invites it back to a URL nobody wants indexed.

The status is site-specific — a theme without that fallback would render an author
archive or 404 — but the reachability is not: it is core's default on every install.

## Why this is not just that site's problem

We model article authors as a post type or a taxonomy on every project, never as
WordPress users. The user is whoever happened to publish. So on our sites an author
archive is a page nobody designed, listing posts grouped by an account name, under a URL
built from a login slug.

## Proposal

A flag, defaulting to on, next to the three that already exist:

```php
/** @var bool Force a 404 on /author/<nicename>/ archives. Default on: this stack models
 *  article authors as a post type or taxonomy, so the core user archive is a page nobody
 *  designed, under a URL built from a login slug. Set false on a site that deliberately
 *  publishes author pages. */
protected bool $disable_author_archives = true;
```

Implemented the same way as `block_author_enumeration` — `set_404()` plus `status_header( 404 )`
on `template_redirect` at priority 9 — so it needs no rewrite flush and takes effect the
moment the flag is read. Removing the rewrite rules instead would be cleaner in principle
and worse in practice: rules are cached in the database, so the flag would appear to do
nothing until someone ran `wp rewrite flush`.

**When it is on, the sitemap exclusion follows automatically.** An archive that 404s must
not be listed anywhere, so `disable_author_sitemap` should be forced on rather than read
independently — the two cannot sensibly disagree.

## The default, and the regression risk

Defaulting to on changes behaviour for existing sites, so I measured the fleet: **26
projects use the kit; 2 route `is_author()`** — `geetoo` and `geetooholding`, both with a
real `@wordpress/author/author.twig` branch. Those two set the flag to `false`. Every
other project gains a 404 where it had an undesigned page.

Defaulting to off would be safe and useless: it is the sites nobody thinks about that
carry this, and they are exactly the ones that would never set it.

Suggest a **minor** release, with the two affected projects named in the notes.

## Open question

Whether SEO-plugin sitemaps should be covered too. AIOSEO and Yoast build their own author
archive entries from their own settings, so on a site running one of them the 404 could
still be listed. The kit already has a plugin-detection registry (`SocialImageBridge` +
`Plugin::active()`), so there is a natural place for it — but it is a second concern and
may belong in its own issue.

parisek:
Closes the fourth username-enumeration vector. Fixes #176.

## What this changes

`disable_author_archives`, default **on**, forces a 404 on `/author/{nicename}/` — on the same hook and priority as `block_author_enumeration`, because it is the same vector.

| Vector | Before | After |
| --- | --- | --- |
| `/?author=N` | 404 | 404 |
| `/wp-json/wp/v2/users` | 401 | 401 |
| `/wp-sitemap-users-1.xml` | 404 | 404 |
| `/author/<nicename>/` | **reachable** | 404 |

`block_author_enumeration` exists because core redirects `/?author=1` to `/author/{username}/` and that redirect leaks the login slug. Its docblock then says path archives are "intentionally untouched" — so the kit blocks the redirect and leaves the destination answering. A crawler reaches the destination without the redirect.

## Why the default is on, and why that is safe

Measured across the **26 projects on this kit**, not assumed:

- **None has a working author archive.** Two route `is_author()` (`geetoo`, `geetooholding`), but neither ships the `author.twig` its branch points at.
- **No theme anywhere links to an author URL.** The `Breadcrumb.php` matches across eight projects are branches that only fire on an author archive — dead once it 404s, harmless.
- One project has an `Author` class; it reads a taxonomy, not users.

A site that deliberately publishes author pages sets the flag to `false`. Defaulting to off would be safe and useless: the sites carrying this are the ones nobody thinks about, and they are exactly the ones that would never set it.

## Two decisions worth reviewing

**It implies `disable_author_sitemap`.** An archive that answers 404 must not be listed anywhere, and a sitemap entry pointing at one is worse than no entry. The two flags are not allowed to disagree.

**404 on `template_redirect`, not removed rewrite rules.** Dropping the rules is cleaner in principle and worse in practice: they are cached in the database, so the flag would appear to do nothing until somebody ran `wp rewrite flush`.

## What made this visible

On `sloneek`, nine author URLs answered **503 with `Retry-After: 3600`**, no `noindex`, a self-referencing canonical, and a real person's name in the `<title>` — a crawler told to come back to a page nobody designed. The 503 is that theme's fail-loud router fallback, so the status is site-specific; the reachability is core's default everywhere.

That is why clearing `is_author` matters as well as setting 404, and it is tested: setting 404 alone lets a theme that routes on `is_author()` render over it.

## Tests

Five new cases, plus one existing test that caught the change before review did — `test_no_hooks_registered_when_all_flags_disabled` fails until the new flag joins its list. Suite green at 1980 tests; PHPStan clean.

## Not in this PR

SEO-plugin sitemaps. AIOSEO and Yoast build their own author entries from their own settings, so on a site running one the 404 could still be listed. The kit has a plugin-detection registry (`SocialImageBridge` + `Plugin::active()`) where that would belong, but it is a separate concern — noted as the open question on #176.


parisek:
## Agent review recorded — Codex (GPT-5), 2026-09-10

Ran against `fd3e6b4`. Six findings, four acted on, one rejected, one already correct. Fixes in `4b7a666`.

| Finding | Verdict | Outcome |
| --- | --- | --- |
| Explicit `is_author = false` before `set_404()` is redundant, and its comment is wrong | **confirmed** | Fixed. Verified in core: `set_404()` calls `init_query_flags()`, which clears every `is_*` flag first. My previous commit message called the line load-bearing; it was not. |
| The test covering it passed only because of a stub deficiency | **confirmed, and the most useful finding** | The suite's `WP_Query` stub implemented `set_404()` as `is_404 = true` and nothing else, so the stub made production code look necessary. Stub now clears flags like core; the test asserts the outcome, not the mechanism. Side effect: two PHP 8.2 dynamic-property deprecations gone from the suite. |
| No assertion that `status_header( 404 )` / `nocache_headers()` are called, or that they are skipped where they must be | **confirmed** | Added. Setting the query flag alone leaves the response a 200 — the exact bug fixed in #169 — and nothing would have caught its return. |
| No assertion on the sitemap coupling | **confirmed** | Added. A `||` quietly becoming `&&` would ship a sitemap advertising 404s with the suite green. |
| `template_redirect` at priority 9 is the right hook and priority | correct as written | No change. |
| Extra `wp_doing_ajax()` / `REST_REQUEST` / `WP_CLI` guards | **rejected** | The review raised these as unconfirmed hardening, not a defect, and said so. `template_redirect` fires only from the frontend template loader behind `wp_using_themes()`, so none of those contexts reaches this code. A guard against a path nothing takes reads as though that path existed. |

The review also flagged the default-on flag as the highest-impact item. That is deliberate and its evidence is in the PR description: across the 26 projects on this kit, none has a working author archive.

Five tests became eight. Suite 1983 green, PHPStan clean.

parisek:
## Second agent review recorded — Codex (GPT-5), 2026-09-10

Ran against the rebased branch. Three findings, all real, all acted on.

| Finding | Verdict | Outcome |
| --- | --- | --- |
| The flag defaults on, and `AGENTS.md` requires new behaviour to default **off** | **confirmed — and it corrected me** | See below. |
| The flag is missing from the README security table | **confirmed** | Added, next to `$block_author_enumeration`, since the two are halves of one guard. |
| No test proves the callback is registered at all | **confirmed** | Added. Every other test calls the method directly, so deleting the `add_action()` line would have shipped a flag that reads as on and does nothing, suite green. The test asserts the hook, that it is registered exactly once, and priority 9 — which is what beats `redirect_canonical`. |

The review also confirmed, with mutation analysis, that the first round's fixes hold: removing `status_header`, removing the `is_admin()` guard, or flipping the sitemap `||` to `&&` are each caught by a named test. It could not run `composer test` / `composer phpstan` (sandbox environment), so those were run here: **1987 green, PHPStan clean.**

### The default, and a bad argument I made

The PR description justified default-on by precedent: `$block_author_enumeration` (1.4.0) and `$disable_author_sitemap` (1.8.0) are both default-on hardening flags. **That precedent does not hold.** The default-off rule landed 2026-06-10 — one day after 1.8.0 and a month after 1.4.0. It was written after them, so citing them reads the rule backwards.

The rule allows an exception the owner reviews explicitly, and names `$seo_canonical_pagination` as the one such case. The owner reviewed this one and approved it on the fleet audit: across the 26 projects on this kit, **none has a working author archive**. It is now recorded in `AGENTS.md` as the second named exception, in the same shape — including the note that the two older flags are not precedent.

PR #179, opened 2026-09-11
feat(seo): keep one BreadcrumbList per page, and make it the theme's

### LINKED ISSUE #178: Every kit project ships two BreadcrumbLists: the shared component's microdata and the SEO plugin's JSON-LD
## What happens

The shared breadcrumb component emits `itemtype="https://schema.org/BreadcrumbList"` microdata. Every SEO plugin we ship alongside it emits its own `BreadcrumbList` node into its JSON-LD `@graph`. Nothing reconciles them, so a crawler is handed two and picks.

## Measured across the fleet

All 25 projects on this kit, checked mechanically:

| | |
| --- | --- |
| Emit BreadcrumbList microdata from the shared component | **25 of 25** |
| Run an SEO plugin (AIOSEO Pro, Yoast, Rank Math) | **25 of 25** |
| Suppress the plugin's copy | **1** — `sloneek`, by hand |

AIOSEO adds it unconditionally: `Schema::getDefaultGraphs()` returns `['BreadcrumbList', 'Kg…', 'WebSite']` with no setting to gate it. So this is not a configuration a project could have got wrong; it is the default everywhere.

## Which copy should win, and why it matters

The theme's. On `sloneek`, Yoast's copy was built from the raw post title and read `Home > A_00_HP_Hybrid` — an English label in a Czech trail, then the editor's internal working title — while the component's microdata on the same page read `Úvod > Hybridně řízené firmy`. That is not a Yoast quirk: the plugin cannot see the trail the theme actually renders, because `timber_kit_breadcrumb_items` and `$breadcrumb_list_page_map` are where that trail is decided.

Giving a crawler the worse of the two, on every page of every project, is the cost.

## The trap in fixing it

Both plugins' `WebPage` node references the list by `@id`. Removing the node alone leaves a dangling `breadcrumb` property pointing at an id that no longer resolves — worse than the duplicate. Yoast needs two filters (`wpseo_schema_graph_pieces` + `wpseo_schema_webpage`); AIOSEO can be done in one pass over the finished graph (`aioseo_schema_output`), which cannot leave the two out of step.

The `sloneek` implementation of both is in `wp-content/themes/sloneek/classes/Base.php` and has been running since August (Yoast) and since the AIOSEO migration (AIOSEO).

## Proposal

`src/Seo/` already carries `Plugin.php`, `Aioseo.php` and `Yoast.php`, so the per-plugin abstraction exists and this is a third method on it rather than new machinery.

A flag — and given #177, **default off**, with the doctrine's own rule applying: this changes rendered output, and a project whose theme does *not* render breadcrumbs would lose the schema entirely by upgrading. `wordpress-base` turns it on, which is where the house default belongs.

```php
/** @var bool Suppress the SEO plugin's own BreadcrumbList so the theme's microdata is the only one. Default off: a theme that renders no breadcrumb would lose the schema entirely. */
protected bool $seo_suppress_plugin_breadcrumb = false;
```

Rank Math is the third case and is not covered by the `sloneek` code — `mk-rolety` runs it. Worth checking whether its filter surface is the same shape before deciding whether the first version covers two plugins or three.

## Open question

Whether the microdata is the right survivor at all, or whether the kit should instead **feed the theme's trail into the plugin** and let the plugin render the JSON-LD. That would keep one source and one output format, and JSON-LD is what Google documents as preferred. It is a bigger change and a different issue; naming it here so the cheap fix is a deliberate choice rather than the only one considered.

parisek:
Fixes #178.

## The problem, measured on two unrelated projects

Every theme on this kit renders a breadcrumb as microdata. The SEO plugin adds a second `BreadcrumbList` to its JSON-LD graph, from data it derived itself, and a crawler is handed two.

The plugin's is the worse copy, because it cannot see the trail the theme built:

| | `rezidence-ponavia` (AIOSEO) | `sloneek` (Yoast) |
| --- | --- | --- |
| page shows | `Úvod > Zahájení prodeje…` | `Úvod > Hybridně řízené firmy` |
| plugin's graph says | `Home > Zahájení prodeje…` | `Home > A_00_HP_Hybrid` |

AIOSEO's `homepageLabel` defaults to `Home` and is not per-language, so a Czech site ships an English root. Yoast's leaf was the editor's internal working title.

## Bound to the plugin's own switch — that is the design

`$seo_suppress_plugin_breadcrumb`, default **on**, but it defers to the plugin's own **Enable Breadcrumbs** toggle. Turn that on and the kit stands aside: an editor who asked for the plugin's breadcrumbs wants its markup too.

That switch is already in wp-admin and already states the intent — it just does not cover the schema today, and AIOSEO's own settings screen says so outright: *"By default AIOSEO will automatically add breadcrumbs to the schema markup that we add to your site."* Binding to it makes the switch mean what it says, and gives every site an escape hatch that needs no code.

**Verified in both directions on a live project:**

| | microdata | JSON-LD | dangling `breadcrumb` ref |
| --- | --- | --- | --- |
| switch **off** (default) | 1 | 0 | 0 |
| switch **on** | 1 | 1 | — |

The pages that render no breadcrumb (`/aktuality/`, `/`) go to 0/0 — AIOSEO had been describing a trail those pages never display.

## The dangling-reference trap

Both plugins' `WebPage` node points at the list by `@id`, so removing the node alone leaves a reference to something that is gone — worse than the duplicate. AIOSEO is filtered on the finished graph (`aioseo_schema_output`), so node and reference go in one pass. Yoast has no equivalent seam and needs both of its filters.

## Two things the repo's own tests corrected

**`SeoBoundaryTest` refused the first shape.** The resolver read the plugins from `StarterBase`, naming an SEO plugin outside `src/Seo/`. The reads moved into the adapters; `StarterBase` now defers to `Seo\BreadcrumbSchema` on `template_redirect`. Better anyway — plugin options are not reliably readable while themes load.

**The runtime caught what no unit test could.** The first read was written `… ?? false`. AIOSEO's `__isset()` is not an existence check: it walks and **resets** the traversal state (`setGroupKey()`, `resetGroups()`). So `??` took a different path and answered `false` while the switch read `true` — silently disabling the escape hatch, with every unit test green, because the decision function takes a boolean and was handed the wrong one.

`AioseoBreadcrumbsEnabledTest` now models that asymmetry and **fails if the operator comes back** — verified by reintroducing it.

## Default-on, and why it is an exception

`AGENTS.md` requires default-off. Recorded there as a named exception, reviewed by the owner. What makes it acceptable rather than merely convenient: the escape hatch is not a code flag but an admin toggle.

The cost is named in the changelog — **8 of 26 projects render no breadcrumb at all** (albaform, anyever, atelier99, fellows, loyaltiq, mairateam, mk-rolety, neoli). Those lose the plugin's list and gain nothing, so they either set the flag `false` or switch the plugin's breadcrumbs on.

## Not covered

Rank Math. Its filter surface was not examined; `mk-rolety` is the project that would need it.

1999 tests green, PHPStan clean.


parisek:
## Review round two — inconclusive, and what I did instead

The second agent review returned a command trace and no findings. Rather than quote it as a clean bill of health, I worked its questions myself. One produced a real change.

**Yoast has a whole-graph seam and I had not used it.** `wpseo_schema_graph` filters the finished array of rendered pieces — the same shape as `aioseo_schema_output`. The branch was using two filters instead: `wpseo_schema_graph_pieces` to drop the generator, `wpseo_schema_webpage` to drop the reference to it.

One filter is strictly better. Two have to agree, and the failure mode when they do not is a `breadcrumb` property pointing at an `@id` no longer in the graph — worse than the duplicate the change exists to remove. One pass cannot reach that state. It also drops a dependency on `Yoast\WP\SEO\Generators\Schema\Breadcrumb`, an internal class name matched by string.

Verified before changing anything: `webpage.php` is the only generator in `src/generators/schema/` that sets `breadcrumb`, so nothing is missed by working a level up. `stripPieces()` and `stripReference()` are gone, and so are their tests — both existed only to coordinate the pair.

### The other questions, answered directly

| Question | Answer |
| --- | --- |
| Is `template_redirect` priority 1 early enough? | Yes. It fires before the template loads, and the schema prints in `wp_head` inside it. |
| Do REST, feeds or sitemaps output schema without `template_redirect`? | No — the schema is head output, and none of those call `wp_head`. |
| Does anything but `WebPage` reference the breadcrumb? | No, verified in the Yoast source. |
| Is the nested-`breadcrumb` case handled? | The property sits at the top level of the node in both plugins; `stripGraph` unsets it there. |

### Re-verified at runtime after the refactor

`rezidence-ponavia`, AIOSEO Pro 4.9.10:

| URL | microdata | JSON-LD | dangling ref |
| --- | --- | --- | --- |
| `/zahajeni-prodeje-bytu-ponavia-rezidence-iv/` | 1 | 0 | 0 |
| `/aktuality/` | 0 | 0 | 0 |
| `/` | 0 | 0 | 0 |

1999 tests green, PHPStan clean, all six checks passing.

PR #180, opened 2026-09-11
fix(helpers): sanitise the announcement text instead of returning it raw

parisek:
## From which project

`neoli`, while migrating the announcement bar to the canonical schema ([neoli#29](https://github.com/portadesign/neoli/pull/29)). That theme had grown a `Base.php` override whose only job was to add the sanitisation this helper does not do:

```php
protected function announcement_context( ?array $value ): array {
    $announcement = Helpers::formatAnnouncement( $value );
    $announcement['text'] = wp_kses_post( $announcement['text'] );
    return $announcement;
}
```

A project working around a helper is usually a sign the helper is missing something.

## Why

`formatAnnouncement()` returns `text` exactly as stored, and the announcement bar renders that field through Alpine's **`x-html`** — so the stored value reaches the DOM as markup.

**Scope, stated precisely** (the first version of this description overstated it — see the correction comment below). `StarterBase` registers `acf/update_value/type=wysiwyg` → `sanitize_acf_editor_value()`, which already runs `wp_kses()` with this same list, and the announcement `text` field is `type=wysiwyg`. **An ordinary save through the ACF UI is therefore already sanitised.**

What is *not* covered is every other write path:

- WP-CLI and SQL — `wp option update`, an import, a DB restore;
- migration scripts, including [neoli#29](https://github.com/portadesign/neoli/pull/29)'s own update file, which writes `options_announcement_text` with `update_option()` and never touches ACF's filter;
- rows written before that hook existed, or on a project not using `StarterBase`;
- revision restore and WPML copy, which move stored values rather than re-saving them through the field type.

So this is **defence in depth on the read side**, not a fix for a live hole. One write path is guarded; the other four are not, and the formatter is the one place they all converge.

It is also the consistency argument: every other editor-content path in `Helpers` already runs `wp_kses()` — `formatLink()` narrows a link title to inline tags, and `getEditorAllowedHtml()` exists precisely to describe what editor richtext may contain. `formatAnnouncement()` simply never used it.

## What it means

```diff
-'text' => $enabled ? (string) ( $value['text'] ?? '' ) : '',
+'text' => $enabled ? wp_kses( (string) ( $value['text'] ?? '' ), self::getEditorAllowedHtml() ) : '',
```

**`getEditorAllowedHtml()` rather than `wp_kses_post()`** — this *is* editor richtext, and the class already defines the list for it, so a consumer reads one rule instead of two. `wp_kses_post()` is also wider than a one-line bar needs. (The downstream override used `wp_kses_post()`; this is the stricter, more consistent choice.)

**No `html_entity_decode()` first**, unlike `formatLink()`. That one decodes because a link field arrives entity-encoded from WordPress; an ACF wysiwyg field does not, and decoding here would turn a stored, harmless `&lt;script&gt;` back into a tag before kses saw it.

## Backward compatibility

For ordinary announcement copy this is a **no-op**, and that is measured rather than assumed — the real stored value on the consuming site is a paragraph with `<strong>` and two `<a href>` links, all of which the editor list passes through byte-for-byte.

Two consequences on a site that stored markup outside the list:

1. that markup is removed;
2. the bar's `sessionStorage` dismissal key is derived from the text, so a changed text forks a new key and a visitor who had dismissed that announcement **sees it once more**.

Both are in the changelog under an upgrade note.

## Tests

Five added, asserting the boundary this class owns — that it calls the right sanitiser with the right list — rather than re-testing WordPress's `kses`. Attribute filtering is kses's own job.

| Test | Asserts |
| --- | --- |
| the editor list is the one passed | `getEditorAllowedHtml()` reaches `wp_kses()` |
| editor markup survives | paragraph + emphasis + link byte-for-byte |
| markup outside the list is removed | `<script>`, `<iframe>` gone, text kept |
| a disabled bar does not sanitise | returns `''`, `wp_kses()` never called |
| a missing `text` key | yields `''` |

**Verified by mutation:** reverting to the raw return fails two of them.

`composer check` clean — phpstan 0 errors, 2004 tests pass.

## Downstream

Once released, `neoli` drops `announcement_context()` and `Base.php` goes back to a one-line call. Any other project that added its own sanitisation can do the same; the changelog says so.

https://claude.ai/code/session_01Jm7pCnFwpyhttR7DZhF5oH


parisek:
## Correction to this PR's threat model — from me, before review

While mapping what else consumes `getEditorAllowedHtml()`, I found that **this PR's original justification overstated the risk**, and the description has been edited.

`StarterBase` registers, unconditionally:

```php
add_filter( 'acf/update_value/type=wysiwyg', array( $this, 'sanitize_acf_editor_value' ), 10, 1 );
```

and the announcement `text` field is `type=wysiwyg`. So an ordinary save through the ACF UI **already** passes `wp_kses()` with this same list. My claim that "a contributor could put a script there" is not true for that path.

### What the change is actually worth

Defence in depth for every write that does **not** go through `acf/update_value`:

- **WP-CLI and SQL** — `wp option update`, an import, a DB restore;
- **migration scripts** — including, concretely, [the one in neoli#29](https://github.com/portadesign/neoli/pull/29), which writes `options_announcement_text` with `update_option()` and never touches ACF's filter. My own change writes unsanitised text into the exact field this PR sanitises on read;
- **rows written before the hook existed**, or on a project not using `StarterBase`;
- **revision restore and WPML copy**, which move stored values around rather than re-saving them through the field type.

Sanitising on read closes all of those at once, and costs nothing when the value was already clean — which the byte-for-byte test covers.

So the change stands, and the reason is narrower and duller than I first wrote: it is not "the bar is exploitable today", it is "one write path is guarded and the other four are not".

I would rather this correction sit on the PR than have a reviewer discover the overstatement themselves.

PR #182, opened 2026-09-15
feat(media): add rescale-originals to apply a raised big image threshold

parisek:
## From which project

neoli raised `$big_image_size_threshold` from the default 2560 to 4000, as fellows did. The 55 attachments already uploaded as `-scaled` stayed at the old size.

## Why

`$big_image_size_threshold` applies at upload only. WordPress keeps the original beside every `-scaled` copy, so a higher threshold can be applied later. Core's `wp media regenerate` reads that original, but it covers only half the cases:

| Original | `wp media regenerate` | `rescale-originals` |
| --- | --- | --- |
| above the threshold | new `-scaled` copy, correct | same |
| fits the threshold | metadata describes the original, `_wp_attached_file` still names the old `-scaled` file, every URL serves it | re-points the attachment, serves the original |

`wp_create_image_subsizes()` calls `update_attached_file()` only when it downscales or rotates, which causes the second row.

Fleet scan of local uploads: 31 projects carry `-scaled` files, most with their originals (ebeton 474, hoteljason 227, pm-a 84, proficio-de 64, …).

## What

- `wp timber-kit rescale-originals [<id>...] [--apply] [--limit=<n>] [--verbose]`, dry-run by default.
- `OriginalImageRescaler` (unit-tested) does the work per attachment:
  - purges resizer derivatives while the attached file still has its old name,
  - points `_wp_attached_file` at the original and runs `wp_create_image_subsizes()`,
  - copies the attached file and metadata to every row over the same file (WPML syncs the first, not the second),
  - restores the previous pointer and metadata when regeneration fails,
  - reports `unchanged` when the threshold is not above the edge the `-scaled` file already has.
- Safety, added during review:
  - every attachment is journalled in post meta (`_timber_kit_rescale_journal`) and its `-scaled` file copied to `.rescale-backup` before the first write; the next run recovers a killed one (dry-run: `interrupted`),
  - success is read back from the database and disk (attached file, served dimensions, every sibling row, `wp_get_missing_image_subsizes()`), never taken from core's return value,
  - the threshold is pinned for the core call; failed rows print a reason.
- `StarterBase::purge_cached_images()` and `attachments_sharing_file()` (protected), passed to the command as closures. `cleanup_cached_images()` behaves as before.
- README (command table, § Raising the threshold for existing images, known limits) and CHANGELOG.

Rejected: a `StarterBase` flag. Nothing runs on upgrade; the command is opt-in by being a command.

## Verification

Three rounds of agent review (Codex, Fable) are summarised in a PR comment.

- `composer test`: 2026 tests green. The one warning (`mkdir(): File exists` in `CleanupCachedImagesTest`) exists on `main` too. `composer phpstan`: no errors.
- neoli (local, branch copied into `vendor/`, backups of uploads and DB taken first): 27 files, 15 restored, 10 rescaled, 2 unchanged; WPML pairs synced. Second run: nothing left to do. Audit of all 154 image attachments: no served file above 4000 px, metadata matches the file on disk, every sub-size exists. 172 image URLs across the site's pages return 200.

Order with `prune-originals`: rescale first. Pruning deletes the originals this command reads.

https://claude.ai/code/session_012E2uU5kuPWuMFtMjGFBuR7 (petr@pari.cz)


parisek:
## Agent review — 2026-09-15

Two independent reviewers, three rounds, different lenses:
- **Codex** (via `codex:codex-rescue`): correctness and data safety against WordPress core source.
- **Claude Fable 5.1**: design, AGENTS.md doctrine, docs accuracy, downstream impact.

Every finding was checked against the code or core source before acting.

### Where they converged
- Round 2: both found that a killed run was not recoverable, contradicting a claim in the round-1 commit message. Led to the journal + backup design.
- Round 3: both flagged filtered metadata reads (`wp_get_attachment_metadata` filter) breaking verification. Fixed with unfiltered reads.

### Fixed
| Round | Finding | Source |
| --- | --- | --- |
| 1 | Core never returns `WP_Error` from `wp_create_image_subsizes()`; a failed rescale reported success | Codex |
| 1 | No rollback on exception | Codex |
| 1 | Dedupe skipped a viable sibling after a non-settling row | Codex, Fable |
| 1 | Non-core metadata keys lost | Codex |
| 1 | Two public `StarterBase` methods without README; made protected, closures passed | Fable |
| 1 | README/CHANGELOG overstated behaviour; one-off measurement in CHANGELOG | Fable |
| 2 | Killed run unrecoverable → post-meta journal + `.rescale-backup` copy, next run recovers | Codex, Fable |
| 2 | Verification trusted the return value → read back attached file, served dimensions, every row, `wp_get_missing_image_subsizes()` | Codex, Fable |
| 2 | Threshold filter evaluated twice → pinned at `PHP_INT_MAX` for the core call | Codex |
| 2 | `unchanged` read from row metadata → measured on the served file | Codex |
| 2 | Sibling writes unchecked → read-back, full rollback | Codex |
| 3 | Journal deleted before verifying; `recover()` could roll back onto overwritten pixels with the backup gone | Codex |
| 3 | Filtered metadata in journal and read-back | Codex, Fable |
| 3 | No failure reason in CLI output | Fable |
| 3 | Partial backup left after failed copy | Codex |

Found by the author while verifying round 3 on a real site: a sibling with a lower ID reported `unchanged`, settled the file and skipped the interrupted row. The command now lists journalled rows first.

### Rejected
- **Round 1: require `$resizer_source_path_in_cache_key`** to avoid legacy basename collisions in the purge. The purge matches `cleanup_cached_images()` and only evicts regenerable derivatives; documented.
- **Round 3, Fable: delete the backup before the journal.** That order leaves a journal without its backup, whose rollback describes pixels no longer on disk. Codex's round-3 finding #1 argued the opposite order, which is what shipped. Recorded because the suggestion looked like a free fix.

### Deferred (documented in README as known limits)
- No lock: two concurrent runs on one attachment are unsafe (Codex r3 #3).
- Recovery overwrites a sibling edited between a killed run and the next (Codex r3 #4).
- Journal stores an absolute path; recover on the same host (Fable r3 #4).

### Reviewer errors
- Round 1, author's own commit claimed "a SIGKILL leaves one attachment for a rerun, which the command handles". False; both reviewers caught it in round 2.
- Round 3, Fable's cleanup-order suggestion would have introduced the divergence Codex described (see Rejected).

### Verification
`composer test` 2051 green (the one warning exists on `main`), PHPStan clean. On neoli (local, from a backup, branch copied into `vendor/`): full run 15 restored / 10 rescaled / 2 unchanged; runs stopped with `exit(137)` during sub-size generation recovered on both IDs of a WPML pair; 154 image attachments audited (none above 4000 px, metadata matches disk, no missing sub-sizes, no journals left); 535 image URLs return 200.

Issue #187, opened 2026-09-15
CLI: wp timber-kit clear-image-cache, for all derivatives or selected images


parisek:
## Problem

Regenerating resizer derivatives today means deleting files by hand on the server (`find wp-content/cache/image -name '*.avif' -delete`). There is no documented command, no dry run, and no way to target one image without knowing the cache layout (flat basename vs source-path key, quality suffixes).

Use cases:
- After a resizer or encoder fix (timber-kit 1.53.0 AVIF quality), regenerate every derivative, or only one format.
- One image was uploaded badly and replaced: regenerate only its derivatives.

## Proposal

`wp timber-kit clear-image-cache [<image>...] [--format=<format>] [--apply]`

- Without `<image>`: every derivative in the resizer cache directory (`timber_kit_resizer_image_cache_dir`).
- `<image>`: an attachment ID, or a file name (`miminko.jpg`, `miminko`). Matches that image's derivatives in both cache layouts, including the `-scaled` file WordPress serves for a large upload.
- `--format=avif`: only derivatives of that format.
- Dry run by default, `--apply` deletes, matching `migrate-image-cache` and `rescale-originals`.
- Never deletes directories or anything outside the cache directory.

Derivatives regenerate on the next request. Pairs with the `timber_kit_resizer_cache_version` issue when browsers also hold old copies.

PR #188, opened 2026-09-15
feat(resizer): add a site-wide cache version to derivative URLs

### LINKED ISSUE #186: Resizer: site-wide cache version appended to variant URLs
## Problem

Resizer derivatives are served from `wp-content/cache/image/` with long browser and proxy lifetimes (neoli.org: `cache-control: max-age=2592000`). A derivative regenerated at the same path keeps its URL, so browsers and CDNs that already hold it serve the old bytes for up to 30 days.

Measured on neoli.org after timber-kit 1.53.0 fixed AVIF quality: the regenerated `1600x0-center-q80/miminko.avif` is 43 KB on the server, but a returning visitor still sees the cached 7 KB file.

A project has no way to force a one-off invalidation of every derivative URL.

## Proposal

- New filter `timber_kit_resizer_cache_version` (string, default empty) and `StarterBase::$resizer_cache_version` wiring.
- When non-empty, Resizer appends `?v=<version>` to every derivative `src` it returns. DevMediaProxy appends the same to the remote variant URLs it builds.
- Paths on disk do not change. Empty keeps today's URLs byte for byte, so no flag.

## Not in scope

Regenerating the files on disk. A version only changes URLs; the server keeps serving the old derivative until it is deleted. See the companion `clear-image-cache` issue.

## Caveats to document

- A CDN set to ignore query strings does not see the change.
- Page cache must be purged so HTML carries the new URLs.

parisek:
Closes #186.

## Why

Derivatives are served with long lifetimes (neoli.org: `cache-control: max-age=2592000`, Varnish in front). Regenerating one at the same path, as after 1.53.0's AVIF quality fix, leaves returning visitors on the old copy for up to 30 days. A project had no way to force a one-off URL change.

## What

- `timber_kit_resizer_cache_version` filter, read and trimmed in the Resizer constructor; `StarterBase::$resizer_cache_version` wires it when non-empty.
- Resizer appends `?v=<version>` (URL-encoded, `&` when a query exists) to every derivative `src` it returns. Passthrough originals are untouched.
- The version travels in the missing-source context; DevMediaProxy probes the bare remote URL and serves the versioned one, so local renders match production.
- `Resizer::appendCacheVersion()` public static, documented in README.
- Tests: URL rule, filter read, context hand-off, proxy URL, StarterBase wiring on/off.

Empty by default leaves URLs byte-identical, so no flag.

## Not in scope

Files on disk. A version changes URLs only; deleting derivatives is #187.

`composer test:all` green (2076 unit, 20 property), PHPStan clean.

https://claude.ai/code/session_012E2uU5kuPWuMFtMjGFBuR7 (petr@pari.cz)


parisek:
## Agent review — 2026-09-15

Codex was unavailable (usage limit); reviewed by **Claude Fable 5.1** across both lenses (correctness and doctrine), two rounds.

**Fixed (round 1):** block cache key did not include the version, so cached blocks served old URLs for up to an hour (now keyed when set, empty keeps keys); the `src` wiring line had no test through `resizer()` (added, verified red on revert); README and a docblock named `clear-image-cache`, which lands in #189; proxy comment and test hygiene.

**Round 2:** nothing material. Checked the other kit caches (MenuFieldsCache, WPML transient, Breeze warmup, SocialImage) — none store derivative URLs.

`composer test:all` green, CI green.

PR #191, opened 2026-09-15
feat(wpml): make the WPML Menus Sync page load read-only

parisek:
## From which project

sloneek (WordPress + WPML 4.9.7). The project guard `classes/MenuSyncReadOnly.php` is the reference implementation.

## Why

WPML → WP Menus Sync writes to the database on a plain page load. WPML hooks `ICLMenusSync::init()` on `init` priority 20. It calls `cleanup_broken_page_items()` and `get_menus_tree()`, which runs `fix_assignment_to_menu()`, `fix_language_conflicts()` and `sync_page_menu_item_trids()` per item.

On 2026-09-14 one visit, closed without confirming, removed 56 items from the cs/sk/pl/it main menus and rewrote 230 `icl_translations` rows. The real sync is a separate AJAX request (`icl_msync_confirm`) that does not run those repairs.

## What

- `Parisek\TimberKit\Wpml\MenuSyncReadOnly`: where WPML sets up menu sync (admin, `page` contains `<ICL_PLUGIN_FOLDER>/menu/menu-sync/menus-sync.php`, case-insensitive, WPML active) it runs `SET autocommit = 0` on `init` priority 1, checks `SELECT @@autocommit`, and runs `ROLLBACK` + `SET autocommit = 1` on `shutdown` priority `PHP_INT_MIN`. Then `wp_cache_flush()` with an external object cache; a failed flush is logged.
- Autocommit, not `START TRANSACTION`: a WPML TM upgrade on `init` priority 10 can send DDL, which commits implicitly. With autocommit off, later writes still roll back.
- Fails closed: `wp_die()` (403) when the `information_schema` query fails, a mandatory table is missing, a table is not InnoDB, autocommit stays on, or the guard registers after `init` priority 1 started (plus `_doing_it_wrong()`).
- `StarterBase::$wpml_menu_sync_read_only`, default off. Hooked on `init` priority 0, or `register()` is called at once when `init` has already fired.
- Scope: writes through global `$wpdb` to InnoDB tables during the request. Not other connections, late shutdown callbacks, sessions or files. Sync reuses a session tree built before the rollback, so it no longer applies WPML's repairs.
- Unit tests for every branch, README section, CHANGELOG entry, ADR 0009.

### Measurements (sloneek, 2026-09-15, before the autocommit switch)

- Guarded page load and preview POST: 0 changed rows.
- SQL general log: one `START TRANSACTION`, one `ROLLBACK`, no `COMMIT`/DDL/`LOCK TABLES` between.
- A confirmed Sync still writes. Other admin pages, front end and admin-ajax open no transaction.
- A fatal error inside the guarded request leaves nothing behind.
- 38 object-cache groups change during the request, hence the full flush.
- A full Sync of 337 items with and without the guard: 0 differing rows in the guarded main menus.

### Rejected alternatives

- Hide the screen: loses the preview.
- Repair data until WPML finds nothing: holds only until the next edit or release.
- Flush only the measured cache groups: the list is one WPML version deep.
- Default on: kit rule is default off.
- Invalidate the session tree after rollback: the unguarded Sync request would rebuild it and persist the repairs.
- Hook `register()` on `after_setup_theme`: a theme that builds `Base` inside that hook would never register the guard.

### Follow-up

- `wordpress-base` sets `$wpml_menu_sync_read_only = true`.
- sloneek replaces its project class with the flag.

https://claude.ai/code/session_013rYAJpapsTdLgiCf8xdCta (petr@pari.cz)


parisek:
## Agent review record

- **Reviewers:** Codex and Claude Opus, 2026-09-15, independent of each other.
- **Verdicts:** both "sound with changes".
- **Where they converged:** the session tree does not match the rolled-back state, and a full cache flush has a wide blast radius.
- **Verification:** the parent session checked each finding against the code and WPML 4.9.7 before any fix.
- **Fixes:** commits 5d2e42f (code and tests) and 2d60c48 (docs).

| Finding | Outcome |
| --- | --- |
| DDL from a WPML TM upgrade (`init` 10) commits the `START TRANSACTION` early | Fixed. The guard uses `SET autocommit = 0`, checks `@@autocommit`, and rolls back, then restores autocommit. |
| Fails open: a failed `information_schema` query counts as "all InnoDB", and the transaction result is ignored | Fixed. Mandatory and optional tables are split. A failed query, a missing mandatory table, a non-InnoDB table or autocommit still on blocks the screen with `wp_die()` (403). |
| Registration during `init` never hooks the guard | Fixed. StarterBase calls `register()` at once after `init`. Past priority 1: `_doing_it_wrong()` and the screen is blocked. |
| Exact `page` match differs from WPML's route check | Fixed. It mirrors `is_core_page()`: case-insensitive substring match, with `ICL_PLUGIN_FOLDER`. |
| `wp_cache_flush()` result ignored; blast radius | Fixed. A failed flush goes to `error_log`. Docs say that multisite or a shared Redis loses the whole cache. |
| Docs claim more than the guard does ("read-only", "nothing survives") | Fixed. Scope is global `$wpdb`, InnoDB, request phase. The session-tree limit and the evaluated alternatives are in ADR 0009. |
| Session tree built before the rollback; Sync reuses it without repairs | Deferred as a documented limit. sloneek measured 0 differing rows in the guarded menus, but that does not prove every data shape. |
| Invalidate `$_SESSION['wpml_menu_sync_menu']` after the rollback, so Sync rebuilds the tree | **Rejected.** |

**Why the session suggestion is rejected, and a reviewer error.** The `icl_msync_confirm` request is not guarded. A rebuild there would run `get_menus_tree()` and its repairs unguarded. It would persist exactly the writes this PR prevents. The suggestion therefore reintroduces the defect. It is recorded in the commit body and in ADR 0009.

Verification after the fixes: `composer test` gives 2133 tests, with only the known Patchwork warning and the WP_DEBUG skip. `composer phpstan` reports no errors.

parisek:
### Verification on a live WPML install, 2026-09-15

The branch was tested on a production copy of a WPML 4.9.7 site (sloneek), with the flag on and the project's own guard removed. Every case has a control run with the guard off, so a zero reads as evidence, not as an absent trigger. Row counts come from a snapshot of menu membership, `icl_translations`, `_menu_item_*` meta, menu items and the location option.

| Case | Guard on | Guard off |
| --- | --- | --- |
| Page load, logged in | 0 rows | 94 rows |
| Preview POST | 0 rows, preview renders | — |
| `CREATE TABLE` on `init` priority 10 | 0 rows, table created | 94 rows with `START TRANSACTION` |
| One table switched to MyISAM | 403, 0 rows | — |
| `information_schema` query forced to fail | 403, 0 rows | — |
| Logged-out request | 0 rows | 94 rows |
| Full Sync of all 337 items | same result as the previous guard, 0 different rows | 0 different rows in the guarded menus |

The SQL general log of a guarded load shows `SET autocommit = 0`, `SELECT @@autocommit`, `ROLLBACK`, `SET autocommit = 1`, and no `START TRANSACTION`.

**New finding, now documented.** A logged-out request is enough. `init` runs before WordPress authenticates, so WPML repairs the menus and only then does `admin.php` redirect to the login form. Fetched with no cookie the URL answers 302 and still changes 94 rows. A crawler that finds the address can break every language. The guard covers it, because `is_admin()` is true there.

**Not covered by this test:** another WPML version, multisite, HyperDB or a read/write split, an object cache other than Object Cache Pro, and a site other than this one. All are named as out of scope in the README.

**Adoption is per project.** The ADR no longer says the `wordpress-base` template turns the flag on.

PR #194, opened 2026-09-16
feat(cli): regenerate the image cache in place instead of deleting it

### LINKED ISSUE #193: Re-encode resizer derivatives in place instead of deleting them
## Problem

The resizer writes derivatives to `wp-content/cache/image/<W>x<H>-<style>[-q<N>]/<source dir>/<source name>.<format>`. Release 1.53.0 fixed AVIF quality: before it, the encoder ignored `timber_kit_resizer_target_quality` and ran at its own default. Every AVIF derivative written before that release therefore carries the wrong encoding. The files are correct-looking and wrong-sized, and nothing on the site reports it.

## Why the existing command is not the answer

`wp timber-kit clear-image-cache` deletes the files and lets them regenerate on the next render. That is unsafe on a live site:

- A page cache keeps serving HTML that points at a file that no longer exists.
- `<source type="image/avif">` inside `<picture>` has no fallback. The browser picks that source and gets a 404, so the visitor sees a broken image, not the JPEG.
- The gap lasts until each page re-renders, which on a large site is hours.

Re-encoding CPU is also a problem on its own. On a multilingual WordPress site with a large media library, regenerating the whole cache through page renders puts the entire encode cost on visitor requests, in an unpredictable order.

## Proposal

`wp timber-kit regenerate-image-cache [<path>...] [--format=<format>] [--older-than=<date>] [--limit=<n>] [--sleep=<ms>] [--max-load=<n>] [--apply]`

Re-encode each derivative in place instead of deleting it:

1. Parse the derivative path back into width, height, style, quality and source file. Every parameter is recoverable from the path.
2. Re-encode through the existing `Resizer` code path into a temp file in the same directory.
3. Verify the temp file decodes and is non-empty.
4. `rename()` it over the target. The rename is atomic on one filesystem, so the URL is never missing or half-written.

A failure at any step keeps the old file. A missing source file is reported as an orphan and never deleted, which stays `clear-image-cache`'s job.

`--older-than` is the resume mechanism: a regenerated file gets a fresh mtime, so the next run skips it. With `--limit`, `--sleep` and `--max-load` a big site spreads the cost over several nights from cron.

## Does it generalise?

Yes. Any kit consumer that upgrades past 1.53.0 has the same wrongly encoded AVIF files, and the same page-cache exposure if it clears them. The same command also covers a future encoder change, a quality change, and a switch of `timber_kit_resizer_target_quality`. It belongs in the kit, not in one theme.

parisek:
Closes #193.

## Why

The resizer writes derivatives to `wp-content/cache/image/<W>x<H>-<style>[-q<N>]/<source dir>/<source name>.<format>`. Release 1.53.0 fixed AVIF quality, so every AVIF derivative written before it is encoded wrong.

`clear-image-cache` deletes them and lets the next render rebuild them. On a live site that is unsafe: a page cache keeps serving HTML pointing at the deleted file, and `<source type="image/avif">` inside `<picture>` has no fallback, so the browser picks that source and the visitor sees a broken image until the page re-renders.

## What

`wp timber-kit regenerate-image-cache [<path>...] [--format=<format>] [--older-than=<date>] [--limit=<n>] [--sleep=<ms>] [--max-load=<n>] [--apply] [--verbose]`

Re-encodes each derivative at its own path:

1. Read the path back into width, height, style, quality, source file and output format.
2. Encode through `Resizer` into a temp file in the same directory.
3. Verify the temp file is non-empty and decodes (`getimagesize()`, then an Imagick ping for AVIF).
4. `rename()` it over the target, which is atomic on one filesystem.

Any failure deletes the temp file and keeps the old one. Nothing is ever deleted: a derivative whose source is gone is reported as an orphan, which stays `clear-image-cache`'s job. Dry-run by default, like `clear-image-cache` and `rescale-originals`.

`--older-than` is the resume mechanism: a regenerated file gets a fresh mtime, so the next run with the same cutoff continues where the last one stopped. With `--limit`, `--sleep` and `--max-load` a large site spreads the encode cost over several nights from cron; the README carries the recipe.

## Shape

- `src/ImageCacheRegenerator.php` — pure, testable logic (parse, resolve, plan, regenerate, load gate), mirroring `ImageCacheCleaner`.
- `src/Cli/RegenerateImageCacheCommand.php` — thin WP_CLI adapter, mirroring `ClearImageCacheCommand`.
- `Resizer::encodeVariant()` — the encode step split out of `processVariant()` so the CLI can aim it at a temp path. Moved, not rewritten, so rendered output is unchanged. Public and `@internal`: the CLI adapter is not in the class hierarchy, so `protected` would have forced a subclass or reflection.

## Limits worth knowing

- The command needs `$resizer_source_path_in_cache_key`. The flat layout names a derivative after the sanitised source stem, which no longer identifies the source file; those are reported as unreadable and left alone. The command warns once rather than refusing, because `migrate-image-cache` moves files one at a time and a cache can be half migrated.
- `smart-crop` re-runs its entropy analysis, so a re-encoded crop may shift a few pixels. Documented.
- The bytes change while the URL does not, so a CDN or browser holding the old copy keeps it. Bump `$resizer_cache_version` when the new bytes must reach everyone at once. Documented.

## After review (Codex and Claude Fable, 2026-09-16)

Both reviews said "safe with changes". The changes are in, and they expand the public surface:

- **Proof before the swap.** The temp file keeps the old file's mode, AVIF is decoded in full rather than pinged, the new file's dimensions must match the old ones, and a new file at or below 1.2x the old size is counted as `suspect` (counted, never refused — a legitimately smaller re-encode exists). The summary prints the median size ratio.
- **New flags.** `--threads=<n>` caps the encoder's threads, `--quality=<n>` overrides the quality a path without `-q` would otherwise inherit from the global filter.
- **One run at a time.** A lock file in the cache directory, released by the kernel. A killed run's temp files are swept on the next start, by PID.
- **Status you can act on.** Non-zero exit when a file failed or when the load gate gave up with work left, a progress bar, a log line when the gate holds, a disk-space preflight and a projection in the dry run.
- **`--older-than` refuses a relative value.** `-2 days` is a moving cutoff that re-encodes the same files every few nights.

Deferred, on purpose: a `--sample --out=<dir>` bench mode that re-encodes a sample into a side directory so an operator can compare qualities before committing. That is a second feature, not a fix.

## Tests

TDD throughout; each suite was watched fail first. 47 new tests: path parsing (including `0x1056`, `-top-`, `smart-crop`, a quality-less directory, a nested source, an encoded/odd filename, `-q0`), refusal of a path outside the cache directory, the `--format`, `--older-than` and `--limit` filters, orphan and unreadable sources, the temp-file failure paths (encoder false, empty file, undecodable file, a throwing encoder) each asserting the original is untouched and no temp is left, the rename success path, dry-run writing nothing, and `sys_getloadavg()` unavailable.

One real defect came out of the tests: `realpath()` follows symlinks, so on a host whose cache directory sits behind one, every resolved path had a prefix the parser did not recognise and nothing was selected. Resolved paths are now rebased onto the configured directory.

`composer test` green (2172 tests, no new warnings — the 17 present are pre-existing, from Patchwork). `composer phpstan` clean on the new files; locally the parallel workers run out of memory on `main` too, so CI is the real check.

No ADR: the command is additive and reversible, and the one decision worth recording is stated in #193, the README and the class docblock.

https://claude.ai/code/session_013rYAJpapsTdLgiCf8xdCta


parisek:
## Review round: Codex and Fable, 2026-09-16

Two read-only agent reviews, independent lenses, both verdicts **"safe with changes"**. Neither is an approve or a request-changes: an agent review is evidence, not a merge gate. Every finding below was checked against the code before it was acted on.

**Where they converged** — four findings both reviewers reached on their own: verification was header-only, stale temp files were never swept, the command exited 0 on a run that failed files, and `--older-than` accepted a relative value that re-encodes the same files every few nights. Convergence between reviewers that cannot see each other's output is the strongest signal here, and all four turned out real.

### Fixed

| # | Finding | What landed |
| --- | --- | --- |
| 1 | File mode and owner | The old file's mode is copied onto the temp file before the rename. The command warns once when the files are owned by another uid. |
| 2 | Verification was header-only | Dimensions of old and new are compared and a mismatch keeps the old file; AVIF goes through a full `readImage()` instead of a ping; each file's size ratio is recorded and a run reports the median and how many came back at or below 1.2x as `suspect`. Suspect is a report, never a refusal. |
| 3 | Encoder threads and load gate | `--threads=<n>` caps Imagick. The README cron recipe now carries `flock`, `nice`, `ionice`, `MAGICK_THREAD_LIMIT`/`OMP_NUM_THREADS` for the binary path, an absolute `wp`, and `--max-load=2` on 4 vCPU with the reason. |
| 4 | Exit status | Non-zero when a file failed or the load gate gave up with work left, printed after the summary. |
| 5 | Run lock | A `flock()` in the cache directory; a second run is refused by name. |
| 6 | Stale temp sweep | Each run deletes temp files whose PID is not alive and reports the count. |
| 7 | `--older-than` semantics | A relative value is refused, detected by reading it against two different "now"s rather than by matching words. Timezone stated in the help and the README. |
| 8 | Quality mapping was silent | `--quality=<n>` overrides it, and every run prints the effective number before it starts. |
| 9 | Progress | A progress bar under `--apply`, and the load gate logs when it starts holding and when it stops. |
| 10 | Disk preflight | Refuses to start under 3x the selected files' weight; the projection prints in the dry run. |
| 11 | README | Carries the 1.53.0 size table, the completion check, the no-op case, and what the checks really prove. |
| 12 | Tests | 42 new tests across the new behaviour, plus the `encodeVariant()` dimension pin the refactor never had. |

### Rejected

Three suggestions inside otherwise-correct findings were not taken, each with the reason in the commit body:

- **Refusing an unmeasurable new file outright** (part of 2). It duplicates the decode check, and it turns a host without Imagick into a host where nothing regenerates. The dimension compare runs only when the old file is measurable — an old file nothing can read is what this command exists to replace.
- **A PID inside the lock file** (part of 5). A run killed with `-9` would leave it behind and lock the cache until someone deleted it by hand, which is the failure the lock exists to prevent. `flock()` is released by the kernel.
- **Deleting every temp file at start** (part of 6). A temp file whose PID is alive, whose name carries no readable PID, or that sits on a platform without `posix_kill()`, is kept: deleting one a live run is writing costs that run its work.

### Deferred, deliberately

A `--sample --out=<dir>` bench mode — re-encode a sample into a side directory to compare qualities before committing to a sweep. Worth doing; it is a second feature, not a fix for anything above, and it belongs in its own PR.

### Checks

`composer test` 2215 tests green, `composer phpstan` clean, all seven CI checks pass.

### One correction worth recording

The 1.2x suspect threshold was proposed as a refusal in one report. Refusing on it would strand exactly the files a legitimately smaller re-encode produces, with nothing to re-encode them afterwards — so it ships as a count in the summary instead. The next reader calibrating how far to trust these reports should know that was the one place the fix as written would have made things worse.

### Context

The site this came from does not need the sweep: its derivatives are already correct, because the fix that mattered is in `spatie/image` 3.9.6, which that site has, and it uses no `smart-crop`. Nothing here was weakened on that basis — the command matters for caches predating that release and for `smart-crop` sites.

https://claude.ai/code/session_013rYAJpapsTdLgiCf8xdCta

PR #199, opened 2026-09-16
feat(social-image): featured-first token, and a bridge that survives an AIOSEO import

### LINKED ISSUE #198: Social preview image: featured-first order, and a bridge that survives an AIOSEO import from Yoast
One case, three defects that only make sense fixed together. Replaces #195, #196, #197.

## The wanted behaviour

The social preview image of a post, in this priority:

1. **the image an editor set in AIOSEO**,
2. **the featured image**,
3. **the mapped ACF field** (`$social_image_fields`),

with `og:image` and `twitter:image` always showing the same picture.

A downstream project needed exactly this. Getting there took a project filter and a data migration. Every project on the bridge would need both.

### 1. The featured image always comes last

`SocialImage::imageCandidates()` ([SocialImage.php#L231-L270](https://github.com/parisek/timber-kit/blob/v1.55.0/src/SocialImage.php#L231-L270)) tries the mapped fields first and the featured image last, with no way to change the order. The downstream workaround bends a filter meant for "read fields from elsewhere" into an ordering switch:

```php
add_filter( 'timber_kit_social_image_post_fields', function ( $fields, $post ) {
	return has_post_thumbnail( $post ) ? array() : $fields;
}, 10, 2 );
```

### 2. After AIOSEO imports from Yoast, the bridge stands aside everywhere, and some posts render no og:image

`defersToEditor()` ([SocialImageBridge.php#L243](https://github.com/parisek/timber-kit/blob/v1.55.0/src/SocialImageBridge.php#L243)) treats any `og_image_type` other than `default` as an editor's choice. AIOSEO's Yoast importer writes a source onto **every** post. Published posts of three mapped types after the import:

| `og_image_type` | posts | featured image | rendered |
| --- | --- | --- | --- |
| `featured` | 1316 | all | featured image |
| `content` | 164 | **none** | **no og:image**: block content yields no image |
| `custom_image` | 165 | — | editor's image |
| `default` | 3 | — | bridge |

Silent: the flag is on, the map is right, the page returns 200, the tag is missing. The downstream project ran a one-off data fix (`content → default`), which every migrated site would have to repeat.

### 3. twitter:image differs from og:image when the editor chose the og image

`filterTwitterTags()` ([SocialImageBridge.php#L189-L209](https://github.com/parisek/timber-kit/blob/v1.55.0/src/SocialImageBridge.php#L189-L209)) decides on `twitter_image_type` alone. With a custom og image, the Twitter source stays `default` and the imported per-post `twitter_use_og = 0` overrides the global "use Facebook data", so `usesOpengraphData()` does not catch it. 147 published posts shared two different pictures. The stopgap was `twitter_use_og = 1` in data.

## Proposal

1. **A `@featured` token in the field map** marks the featured image's position: `'post' => array( '@featured', 'image' )`. Without it the order is unchanged, so no existing configuration moves.
2. **Defer only to an editor's actual image:** `custom_image` and `custom` (custom-field source). For `featured` and `content`, supply the preview. For `featured` the resolved candidate is the featured image anyway when the map puts `@featured` first, so a post that has one keeps it.
3. **`twitter:image` follows the og decision:** when the og source defers to the editor, the Twitter filter stands aside too (AIOSEO renders the editor's image for both).

With all three, the downstream filter and data migration are deleted.

## Does it hold across projects

Of 112 local WordPress projects, **104 run All in One SEO** and 26 use timber-kit. The base skeleton and one more project enable the bridge with a single mapped field, so both get "field first" today. Every site migrated from Yoast carries imported sources on every post.

## Rejected

- **Flipping the default order to featured-first.** It would change og:image on every mapped post of every project that already enabled the bridge.
- **A boolean `$social_image_prefer_featured`.** It cannot express a per-type order; the map already is the per-type setting.
- **Treating every non-custom source as `default`, including `featured` without a map token.** On a post with a featured image that would replace a working image with the field.
- **Documenting a data migration instead.** It must be re-run per site, and a new import brings the problem back.

parisek:
Closes #198.

## What

One wanted priority for the social preview image: **the editor's AIOSEO image > the featured image > the mapped field**, with `og:image` and `twitter:image` on the same picture.

1. **`SocialImage::FEATURED` (`'@featured'`)** in `$social_image_fields` places the featured image in a chain: `'post' => array( '@featured', 'image' )`. Without the token the featured image stays last. Fields are read only when the chain names one.
2. **`defersToEditor()`** is now `custom_image` and `custom` only. For the automatic sources:
   - `featured` keeps the featured image; the bridge supplies only where AIOSEO fell back to its global default image or the site logo.
   - `content`, `attach`, `author` and `auth` pick an arbitrary body image, attachment or avatar, so the bridge supplies wherever a preview resolves.
   - An unknown source still defers.

   The decision is the pure `shouldSupply()` / `isPluginFallback()`.
3. **`filterTwitterTags()`** copies the resolved Open Graph image (`aioseo()->social->facebook->getImage()`, which runs the og filter) instead of cutting a separate preview. That is AIOSEO's own Twitter fallback, and it keeps both cards on one picture.

## Behaviour change: owner review

`$social_image_bridge` is already opt-in. Changes 2 and 3 alter output **for sites that already set it**:

- a missing or fallback `og:image` gets filled;
- a mismatched `twitter:image` now matches `og:image`.

Both are listed under `### Changed` in the changelog. AGENTS.md asks for new behaviour behind a flag; I did not add a second flag, because the old behaviour is the defect in both cases. That is the owner's call.

## Verified downstream

On a production copy of a project on 1.55, with AIOSEO imported from Yoast. The branch's two source files were dropped into its vendor. The theme map was `array( '@featured', 'article_image' )`, with **no theme filter and no data migration**. 13 posts, each on its own language URL, with the rendered post id checked:

| Source | Featured image | Before | With this branch |
| --- | --- | --- | --- |
| `featured` (5 posts) | yes | featured / featured | unchanged |
| `content` (3 posts) | no | **no og:image** (2), CTA banner from body (1) | mapped field, og = twitter |
| `custom_image` (2 posts) | yes | editor's image | editor's image, og = twitter |
| `default` (3 posts) | mixed | none | featured if any, else field |

This matches what the project had reached with a theme filter plus a data migration, which it can now delete.

The first version of rule 2 kept whatever `content` found. The same measurement caught the CTA banner, and the second commit fixes it.

## Checks

- `composer test`: 2225 tests OK. Warnings 17 and skipped 5 are identical on `main`.
- `composer test:property`: OK.
- `composer phpstan`: no errors.
- `composer adr`: OK.

New tests:
- `ForPostTest`: featured token first, fall-through without a thumbnail, token never read as a field.
- `FieldNamesTest`: token keeps its position.
- `BridgeTest`: editor vs automatic sources, `isPluginFallback()`, `shouldSupply()` per source, Twitter follows og, Twitter untouched when og is unreadable.

`BridgeTest::test_an_explicit_per_post_choice_is_left_alone` used to assert that `featured` defers. That assertion is replaced, because it encoded the defect.

## Rejected

- **Flipping the default order to featured-first.** It moves `og:image` on every site already on the bridge.
- **A boolean `$social_image_prefer_featured`.** It cannot express a per-type order.
- **Treating `featured` like `default` without the token.** It replaces a working featured image with the field.
- **A static test seam for the preview resolver.** The decision is pure and tested directly instead.


parisek:
## Agent review: Codex (via Claude Code), 2026-09-16

An independent Codex review ran on the branch, read-only. I checked each finding against AIOSEO Pro's source before acting.

| # | Finding (Codex severity) | Outcome |
| --- | --- | --- |
| 1 | A post left on `default` ignores AIOSEO's **global** source. With a global `featured`, the mapped field replaced the featured image (High) | **Fixed.** Confirmed in `Facebook::getImage()`. New pure `effectiveSource()`, with tests. Note: the old kit behaved the same way; this is not a regression of the PR. |
| 2 | Output changes for sites already on `$social_image_bridge` without a new default-off flag (High) | **Open, owner decision.** Either a named exception in `AGENTS.md` (the old behaviour is the defect: a missing or mismatched tag) or a compatibility flag. |
| 3 | Under `featured`, AIOSEO returns a persisted `og_image_url` before reading the thumbnail, so a stale cached URL counts as "found" (Medium) | **Deferred.** The mechanism is confirmed in `Image::getFeaturedImage()`; the impact is not reproduced. No change without a reproduction. |
| 4 | A Twitter image the editor set separately can differ from og (Medium) | **Rejected as a defect.** It is an explicit editor choice, which the contract never overrides. The README now says so. |
| 5 | `auth` is not an AIOSEO source (Low) | **Fixed.** Codex was right about `auth`, but **missed `auto`**. `auto` is a real source ("first available image") that the branch deferred as unknown. Both are corrected. |

**Checked by Codex and not a problem:**
- **Recursion:** the Twitter filter re-enters the og filter, but the og filter does not call back.
- **AIOSEO's hash cache:** the key includes the source and the post.
- **The homepage and archives:** they never reach the bridge.
- **Unknown sources:** they defer.

Codex noted that Twitter and the size lookup do repeated preview work, which costs time but not correctness.

**Test gaps it listed, not addressed here:**
- integration against real AIOSEO objects;
- URL normalisation;
- WPML metadata.

The downstream check covers part of that.

**After the fixes:**
- `composer test`: 2227 OK. Warnings and skips are identical to `main`.
- `phpstan`: no errors.
- Downstream re-run on the same 13 posts gave a result identical to the previous verified run, with `og:image` = `twitter:image` on all.

parisek:
## The flag question, resolved as an exception

`AGENTS.md` now carries a named exception for this change (latest commit), in the shape the four existing ones use. In short:

- The rule governs **new behaviour arriving behind a flag**. This is a change to behaviour a consumer had already opted into with `$social_image_bridge`.
- **The old behaviour is the defect in both halves**: an imported post with source `content` rendered no `og:image` on block content, and `twitter:image` was resolved separately from `og:image`. Measured downstream: 164 posts with no image, 147 with two different pictures.
- A default-off compatibility flag would leave both running on every site that did not flip it, and nobody flips a flag for a tag they cannot see. Same reasoning as `$disable_author_archives` and `Helpers::getEditorAllowedHtml()`.
- **What bounds it:** an image the editor chose in the plugin's panel (`custom_image`, `custom`) is never touched, a source the code cannot name defers, and a site that wants none of it turns the bridge off — the flag it already has.

Merging the PR is the explicit review the exception records.

parisek:
## Codex review, four rounds

`codex-cli` 0.151.0, gpt-5.1-codex (high reasoning), 2026-09-17. Four rounds, each one reviewing the previous round's fix rather than the whole branch again. Summarised, not pasted.

### Round 1 — three findings, all real, all fixed

| Finding | Outcome |
| --- | --- |
| The Twitter path read `twitter_image_type` alone, so a site whose global Twitter source is `custom` had every post left on `default` overwritten — an editor's card image, replaced | **fixed** — both paths now decide on the effective source |
| An unknown Twitter source was overwritten instead of deferred, contradicting README, CHANGELOG and AGENTS.md | **fixed** — one `shouldSupply()` for both networks |
| `featured` compared the resolved URL against AIOSEO's fallbacks, so a featured image that is also the site's default social image read as a fallback and was replaced | **fixed** — the decision asks the post; `isPluginFallback()` and `pluginFallbacks()` are gone, and with them one walk through another plugin's internals |

### Round 2 — the fix had introduced a regression

`has_post_thumbnail()` answers whether an ID is assigned, not whether the post has a picture. An ID whose attachment is gone still answers yes, AIOSEO falls back to its default image, and the bridge stood aside for an image that is not the post's — this branch's own defect, in a narrower case. **Fixed**: `hasFeaturedImage()` resolves the attachment URL.

The same round showed **both new Twitter tests passed for the wrong reason**. With no Facebook resolver stubbed, the filter returned the meta unchanged for want of an image to copy, so an implementation that always proceeded satisfied them too. **Fixed**, and verified by mutation.

### Round 3 — two small ones

`hasFeaturedImage()` ran for every source although only `featured` reads it, paying for an attachment lookup and the image-downsize filters to answer a question nobody asked — gated by `featuredImageState()`. And the two *older* Twitter deferral tests had the same weakness as the new ones — stubbed.

### Round 4 — merge

> Merge — both Round 3 issues are fixed, and commit `6601733` introduces no new defects.

### Rejected, and why

- **A test proving `getMetaData()` is called exactly once.** The single read is four lines of one method, and a call-count assertion pins the implementation rather than the behaviour.
- **Integration tests against real AIOSEO and WordPress** (round 1 asked for none; the brief ruled them out). The package runs on stubs deliberately. AIOSEO's persisted `og_image_url` cache, URL scheme differences and WPML belong in a downstream behaviour contract, where a database and a real plugin exist.

### Where the reviewer was wrong

Nowhere material across four rounds — each finding reproduced in the code. One observation was overstated rather than wrong: round 3 called the eager `hasFeaturedImage()` a cost "unlikely to be a major database cost", which is right, and it was worth fixing for clarity more than for speed.

### What the review is worth knowing for

Two defects were in production code and **three were in tests that passed without proving anything** — an assertion satisfiable by a path the test does not watch. Every new test on this branch had that shape on the first attempt. Each one is now mutation-checked: a bridge that never defers on Twitter kills four tests, one that ignores "Use Data from Facebook Tab" kills a fifth, and one that trusts an assigned thumbnail ID kills the dangling-ID test. None of those mutants died before the fixes.

https://claude.ai/code/session_01V5w3CibziJQLBS7XwHEmk1

